PR #405, opened 2020-03-24
B/background scheduler tests

Alexander-Dubrawski:
Add test for background scheduler.

## Todo

- [x] Add last tests
- [x] Refactor the way we access patched mocks
- [x] Add connsistent typing 
- [x] Check out `mock_calls` 
- [x] Handle not existing tables in `_get_load_table_names` (`background scheduler`)


tests/database_manager/test_background_scheduler.py
```diff
@@ -0,0 +1,223 @@
...
+            (background_job_manager._update_storage_data, "interval", 5),
+            (background_job_manager._update_plugin_log, "interval", 1),
+        ]
+        for func, _, _ in jobs:
+            func = MagicMock()
+
+        background_job_manager._init_jobs()
+
+        for func, trigger, seconds in jobs:
+            mock_scheduler.add_job.assert_any_call(
```
fabianhe: Why not `assert_called_once_with`?

Alexander-Dubrawski: The docs say to `assert_called_once_with`:
Assert that the mock was called exactly once and that that call was with the specified arguments.

https://docs.python.org/3/library/unittest.mock.html

Since we are calling this function more than one-time witch different arguments we are using `assert_any_call`.  🤓

fabianhe: True, nevertheless I suggest using a more powerful assertion than `assert_any_call`.
E.g. you could use `mock_calls` (which contains all calls made to a method) in the loop to assert that the expected calls and no other calls were made.

tests/database_manager/test_background_scheduler.py
```diff
@@ -35,17 +37,24 @@
 )
 
 mocked_storage_curser = MagicMock()
+mocked_pool_curser = MagicMock()
```
fabianhe:
```suggestion
mocked_pool_cursor = MagicMock()
```

tests/database_manager/test_background_scheduler.py
```diff
@@ -0,0 +1,1183 @@
...
+            "Tenacious D": "Rock",
+        }
+
+        background_job_manager._loaded_tables = fake_loaded_tables
+
+        received: List = background_job_manager._get_load_table_names(
+            "techno techno... :("
+        )
+
+        assert not received
```
PeterTsayun: `False`, `None` and `[]` return `False` when converted to boolean. So, I think, it can be useful to distinguish between them, like:
```
assert received == []
```

tests/database_manager/test_background_scheduler.py
```diff
@@ -0,0 +1,1183 @@
...
+        )
+
+        global mocked_process
+        global mocked_process_constructor
+
+        args, kwargs = mocked_process_constructor.call_args_list[0]
+
+        assert kwargs["target"]() == 42
+        assert kwargs["args"][0] == fake_queries[0]
+        assert type(kwargs["args"][1]) is ValueType
```
fabianhe:
```suggestion
        assert isinstance(kwargs["args"][1], ValueType)
```

PR #407, opened 2020-03-24
f/400-setup-frontend-tests

### LINKED ISSUE #400: setup frontend tests
Currently there is no test coverage for frontend code.

We are not gonna test if every html element is on the right place and has the correct color, but we still need some basic assertions to ensure we are not adding bugs with new features. Thats why we need to setup some tests.

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #400

This PR adds a setup for frontend e2e tests with cypress. Additionally there will be some first routing tests.

**This PR adds:**
- setup for cypress as testing framework ([docs](https://docs.cypress.io/guides/overview/why-cypress.html#In-a-nutshell))
- basic routing tests
- setup for mocked backend requests
- blueprint and possibility of how to handle post / delete requests

**Affected Component(s):**  
`Frontend`


hyrisecockpit/frontend/src/components/details/AvailableDatabasesList.vue
```diff
@@ -1,12 +1,13 @@
 <template>
-  <v-list class="list">
+  <v-list id="database-list" class="list">
```
MaxSchneider1337: since I see so many ids, should we add an id by default in the future?

MaxSchneider1337: we should probably add some conventions for id-naming so we can guess the ids without looking in the code.

schTi: Yes we should definitely add ids to relevant html elements, to make testing easier

schTi:
NOTE: see #414

PR #409, opened 2020-03-25
B/database manager tests

PeterTsayun:
# Resolves #number

This PR adds `DatabaseManager` tests and increases the test coverage of that component up to 94%.

No feature added.

**Affected Component(s):**  
`DatabaseManager`

**Additional context:**


tests/database_manager/test_database_manager.py
```diff
@@ -134,3 +167,478 @@ def test_call_queue_length_returns_queue_length(
...
+
+    def test_delete_database(self, database_manager: DatabaseManager) -> None:
+        """Test delete database."""
+        database = fake_database()
+        database_manager._databases["db1"] = database
+        body: Dict = {"id": "db1"}
+        response = database_manager._call_delete_database(body)
+
+        assert get_response(200) == response
+        assert list(database_manager._databases.values()) == []
```
Alexander-Dubrawski: I think we should check if the `_databases` dictionary has no entry under the key `db1`.

```suggestion
        assert not "db1" in database_manager._databases
```

PeterTsayun: Probably, `not "db1" in database_manager._databases.keys()` :smirk:

tests/database_manager/test_database_manager.py
```diff
@@ -1,28 +1,54 @@
...
 class TestDatabaseManager:
     """Tests for the DatabaseManager class."""
 
     @fixture
+    @patch(
+        "hyrisecockpit.database_manager.manager.Server", fake_server_constructor,
+    )
     def database_manager(self) -> DatabaseManager:
         """Get a new DatabaseManager."""
         with DatabaseManager(
```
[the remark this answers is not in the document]
PeterTsayun: But in that way, we can cover the context manager methods as well :grin:

hyrisecockpit/database_manager/manager.py
```diff
@@ -43,7 +43,29 @@ def __init__(
         self._storage_user = storage_user
 
         self._databases: Dict[str, Database] = {}
-        server_calls: Dict[str, Tuple[Callable[[Body], Response], Optional[Dict]]] = {
+        server_calls: Dict[
+            str, Tuple[Callable[[Body], Response], Optional[Dict]]
+        ] = self._get_server_calls()
```
fabianhe: Why move this to a method if it's static?

tests/database_manager/test_database_manager.py
```diff
@@ -1,28 +1,54 @@
...
+
+def fake_server_constructor(*args) -> MagicMock:
+    """Fake server."""
+    fake_server = MagicMock()
+    fake_server.start.return_value = None
+    fake_server.close.return_value = None
+
+    fake_server_constructor = MagicMock()
+    fake_server_constructor.return_value = fake_server
+    return fake_server_constructor
```
fabianhe: How about using `patch.object` and patching `start` and `close`?

tests/database_manager/test_database_manager.py
```diff
@@ -1,28 +1,54 @@
...
+def fake_database(*args) -> MagicMock:
+    """Fake server."""
+    fake_database = MagicMock()
+    fake_database.driver.dbname = "database_name"
+    fake_database.driver.host = "database_host"
+    fake_database.driver.port = 10000
+    fake_database.number_workers = 8
+    fake_database.close.return_value = None
+
+    return fake_database
```
fabianhe: How about using `patch.object` and patching `close`?

tests/database_manager/test_database_manager.py
```diff
@@ -134,3 +167,492 @@ def test_call_queue_length_returns_queue_length(
...
+            "activate plugin",
+            "deactivate plugin",
+            "set plugin setting",
+            "get plugin setting",
+        ],
+    )
+    def test_has_server_call(
+        self, database_manager: DatabaseManager, call: str
+    ) -> None:
+        """Assert DatabaseMnager has a specific call."""
```
fabianhe:
```suggestion
        """Assert DatabaseManager has a specific call."""
```

tests/database_manager/test_database_manager.py
```diff
@@ -134,3 +167,492 @@ def test_call_queue_length_returns_queue_length(
...
+        assert call in database_manager._get_server_calls().keys()
```
fabianhe: This test will not fail if the calls change by adding a new one to the `DatabaseManager`, how about we compare the calls listed here with the calls the `DatabaseManager` has for equality?

hyrisecockpit/database_manager/manager.py
PeterTsayun: In my opinion, it is more readable and actions in the `__init__`-method stay roughly at the same level of abstraction. Moreover, it is easier to test

tests/database_manager/test_database_manager.py
PeterTsayun: Which advantage would it have?

tests/database_manager/test_database_manager.py
PeterTsayun: Good point, thanks for the suggestion

PR #416, opened 2020-03-26
f/411-add-timestamp-interval

### LINKED ISSUE #411: Show request interval in access map
The user does not know how often the accessmap data is fetched. 

Show the interval in the card (use the whitespace)

could also be shown on element-hover, but this seems to be much harder and hard to find for the user

**Affected Component(s):**  
frontend

schTi:
# Resolves #411

Currently nobody knows of which time interval the displayed data is.

This PR adds a time interval for every metric which is not shown in a line chart, which already has time stamps.

**Affected Component(s):**  
`Frontend`

**Additional Context**
image

### DIFF
--- a/hyrisecockpit/frontend/src/meta/formatting.ts
+++ b/hyrisecockpit/frontend/src/meta/formatting.ts
@@ -29,7 +29,7 @@ export function useFormatting(): {
   }
 
   function formatDateToHHMMSS(date: Date): string {
-    return date.toLocaleTimeString();
+    return date.toLocaleTimeString("de-DE");
   }
 
   function roundNumber(
--- a/hyrisecockpit/frontend/src/meta/formatting.ts
+++ b/hyrisecockpit/frontend/src/meta/formatting.ts
@@ -7,6 +7,7 @@ export function useFormatting(): {
     factor?: number,
     even?: boolean
   ) => number;
+  formatDateToHHMMSS: (date: Date) => string;
 } {
   function formatNumberWithCommas(data: number): string {
     const parts = data.toString().split(".");
@@ -27,6 +28,10 @@ export function useFormatting(): {
     return date;
   }
 
+  function formatDateToHHMMSS(date: Date): string {
+    return date.toLocaleTimeString();
+  }
+
   function roundNumber(
     data: number,
     ratio: number,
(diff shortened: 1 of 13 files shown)


schTi:
image

hyrisecockpit/frontend/src/meta/formatting.ts
```diff
@@ -27,6 +28,10 @@ export function useFormatting(): {
     return date;
   }
 
+  function formatDateToHHMMSS(date: Date): string {
+    return date.toLocaleTimeString();
```
MaxSchneider1337: Why did you put this in an extra function? since toLocalString is a default js function it should be understandable enough.

MaxSchneider1337: we should also think about a locale for our dates, now the browsers locale is used, which could cause some issues when the system is used in the US

hyrisecockpit/frontend/src/views/AppDrawer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-navigation-drawer v-model="open" app width="110">
```
MaxSchneider1337: good job, just saw the error too

schTi: Yes this triggered me all the time

hyrisecockpit/frontend/src/meta/formatting.ts
schTi: Thats exactly the reason why I used an extra function. Im not sure if we should use the local time string function. Du you have a good idea to solve this without this function?

schTi: What do you think about adding date-fns as dependency? Its a nice package for handling dates. Maybe we need some more date handling in the future? [](https://date-fns.org/v2.11.0/docs/format)

MaxSchneider1337: I think it would be overkill to use an extra lib right now. I would suggest using the german locale for every timestamp without a date.

schTi: Ok Im fine with that

PR #420, opened 2020-03-26
f/418-add-possibility-to-remove-dbs

### LINKED ISSUE #418: add possibility to remove databases
Currently it is not possible to remove a already registered database.

We should add a way to make this possible. Just use the already available endpoint for this. Maybe we should show a warning, which has to be accepted before deleting the database.

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #418

This PR adds the possibility to remove databases via a dialog.

**Affected Component(s):**  
`Frontend`

**Additional context:**  
database list:

image

remove db dialog:
image

### DIFF
--- a/hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
+++ b/hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
@@ -1,0 +1,73 @@
+<template>
+  <v-dialog v-model="open" persistent max-width="400px">
+    <v-card id="add-database">
+      <v-card-title>
+        <span class="headline">Remove database</span>
+      </v-card-title>
+      <v-card-text>
+        <div>Do you really want to remove this database?</div>
+        <database-chip class="mt-4" :database-id="databaseId"
+      /></v-card-text>
+
+      <v-card-actions>
+        <v-spacer></v-spacer>
+        <v-btn color="primary" text @click="$emit('close')">Cancel</v-btn>
+        <v-btn
+          id="save-database-button"
+          color="primary"
+          text
+          @click="removeDatabase()"
+          >Remove</v-btn
+        >
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import {
+  defineComponent,
+  SetupContext,
+  onMounted,
+  computed,
+  Ref,
+  ref,
+  watch
+} from "@vue/composition-api";
+import { useDatabaseService } from "@/services/databaseService";
+import DatabaseChip from "@/components/details/DatabaseChip.vue";
+
+interface Props {
+  open: boolean;
+  databaseId: string;
+}
+interface Data {
+  removeDatabase: () => void;
+}
+
+export default defineComponent({
+  props: {
+    open: {
+      type: Boolean,
+      default: null
+    },
+    databaseId: {
+      type: String,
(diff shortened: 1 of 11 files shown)


hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
```diff
@@ -0,0 +1,73 @@
...
+      </v-card-title>
+      <v-card-text>
+        <div>Do you really want to remove this database?</div>
+        <database-chip class="mt-4" :database-id="databaseId" />
+      </v-card-text>
+      <v-card-actions>
+        <v-spacer />
+        <v-btn color="primary" text @click="$emit('close')">Cancel</v-btn>
+        <v-btn
+          id="save-database-button"
```
MaxSchneider1337: should this be a remove-database-button

hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
```diff
@@ -0,0 +1,73 @@
...
+    databaseId: {
+      type: String,
+      default: null
+    }
+  },
+  components: {
+    DatabaseChip
+  },
+  setup(props: Props, context: SetupContext): Data {
+    const databaseService = useDatabaseService();
```
MaxSchneider1337: why do you use the whole databaseService here and not just use` const {removeDatabase} = useDatabaseService();`

MaxSchneider1337: a ok is see, its because of the function name, then i would probably rename the components function to something like `onRemoveDatabaseButtonClick`.

hyrisecockpit/frontend/src/services/databaseService.ts
```diff
@@ -100,8 +103,24 @@ export function useDatabaseService(): DatabaseService {
     });
   }
 
+  function removeDatabase(databaseId: string): void {
+    axios
+      .delete(controlBackend + "database", {
+        data: { id: databaseId }
+      })
+      .then(() => {
+        emitDatabaseRemovedEvent();
```
MaxSchneider1337: I think, that the pluginService should refetch the plugins and settings on this event. Should work without it as well, but should be cleaner.

hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
schTi: good catch, thanks

hyrisecockpit/frontend/src/components/databases/RemoveDatabase.vue
schTi: To be honest I just prefered this synatx, because its a bit more class-styling. Imo the dabase service capsules all database functionality, and it is more convenient for me to just call some of his functionality directly. So I dont need to import the functionality step by step.

hyrisecockpit/frontend/src/services/databaseService.ts
schTi: Good point. I think we should definitely refactor the plugin service as well. It contradicts with our coding conventions, because we should split the service into controller and service as well.

schTi: thats what I just thought about

PR #452, opened 2020-03-30
f/451-simplify-db-editor

### LINKED ISSUE #451: simplify addDatabase-Editor
To add a Database the user has to fill in many textfields that are not needed.

Let the user only fill host and workers (id could be auto-genereated in backend). rest of fields could be hidden behind a "expert"-view

**Affected Component(s):**  
Frontend, AddDatabseEditor

MaxSchneider1337:
# Resolves #451

complex add db editor

makes it simpler

**Affected Component(s):**  
addDatabse

**Additional context:**  
image

image

### DIFF
--- a/hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
+++ b/hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
@@ -27,9 +27,8 @@
                 required
               ></v-text-field>
             </v-col>
-            <spacer/>
+            <v-spacer/>
             <v-btn
-              class="advanced-btn"
               text
               @click="showAdvanced = !showAdvanced"
             >
@@ -210,8 +209,4 @@ function useDatabaseCreation(context: SetupContext): DatabaseCreationData {
 }
 </script>
 <style scoped>
-.advanced-btn {
-  position: absolute;
-  right: 24px;
-}
 </style>
--- a/hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
+++ b/hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
@@ -15,7 +15,12 @@
               ></v-text-field>
             </v-col>
             <v-col cols="6">
-              <v-text-field v-model="id" label="Id*" required></v-text-field>
+              <v-text-field
+                v-model="id"
+                label="Id*"
+                required
+                :error-messages="idError"
+              ></v-text-field>
             </v-col>
           </v-row>
           <v-row align="center">
@@ -96,6 +101,7 @@
             showAdvanced = false;
             closeDialog();
           "
+          :disabled="!!idError.length"
           >Save</v-btn
         >
       </v-card-actions>
@@ -146,6 +152,7 @@ interface DatabaseCreationData {
   host: Ref<string>;
   port: Ref<string>;
   dbname: Ref<string>;
+  idError: Ref<string>;
   createNewDatabase: () => void;
   closeDialog: () => void;
 }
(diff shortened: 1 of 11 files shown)


hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
```diff
@@ -51,22 +27,72 @@
                 required
               ></v-text-field>
             </v-col>
-            <v-col cols="12" sm="6">
-              <v-text-field v-model="id" label="Id*" required></v-text-field>
-            </v-col>
           </v-row>
+          <v-expand-transition>
```
schTi: Maybe use an expansion panel for this, then you have a built-in expand icon

hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
```diff
@@ -130,6 +160,10 @@ function useDatabaseCreation(context: SetupContext): DatabaseCreationData {
   const port = ref<string>("5432");
   const dbname = ref<string>("postgres");
 
+  watch(host, host => {
```
schTi: ID should be a computed, instead of watch. And then you should make the ID a readonly field

MaxSchneider1337: ID shoud stay like this, because it should be the same as host as long as I don't input in id directly. 

So for example, when the host id is already taken i can add some suffixes

hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
MaxSchneider1337: naa really don't like it, cause it is manly build for multiple panels and sometimes breaks when used with just one panel

MaxSchneider1337: over all i ended up having less code implementing it like this.

schTi: Im fine with that, but please make the "advanced" button fixed on the top. What I means is, it should stay on top of the advanced settings if they were open, and not move to the bottom.

hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
schTi: but when you add host -->then add id --> then adapt host --> id will be the same as host

schTi: you should only set id=host, when id is empty

MaxSchneider1337: or when host === id

schTi: this seems to be trivial :smile:

hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
MaxSchneider1337: test

MaxSchneider1337:
@schTi should be ready

hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
```diff
@@ -159,9 +167,20 @@ function useDatabaseCreation(context: SetupContext): DatabaseCreationData {
...
   const dbname = ref<string>("postgres");
+  const idError = ref<string>("");
 
-  watch(host, host => {
-    id.value = host;
+  watch(host, (host, prevHost) => {
+    if (!id.value.length || id.value === prevHost) id.value = host;
+  });
+
+  watch(id, id => {
```
schTi: I really like the error handling

PR #461, opened 2020-03-30
B/handle hyrise failure

Alexander-Dubrawski:
This PR is implementing error handling for the case that the hyrise is down. We are now catching the *psycopg2* `DatabaseErrors` and  `InterfaceErrors`.

https://www.psycopg.org/docs/errors.html

The error handling is directly at the components.  As a result, the components can act individualized if the hyrise is down. A disadvantage is that the components now need to catch the exception. 
The `_ping_hyrise_job` in the `BackgroundJobManager` is checking every half-second if the hyrise is still alive.

If we call the endpoint `monitor/status` we can check the status of the hyrise database: 

```json
[
    {
        "id": "db2",
        "hyrise_active": true,
        "database_blocked_status": false,
        "worker_pool_status": "closed",
        "loaded_benchmarks": [
            "tpch_0.1"
        ],
        "loaded_tables": [
            {
                "table_name": "customer",
                "benchmark": "tpch_0.1"
            },
            {
                "table_name": "lineitem",
                "benchmark": "tpch_0.1"
            },
        ]
    }
]
```

**Affected Component(s):**  
`Cursor`, `BackgriundJObScheduler`

It is also possible to just simply restart the hyrise. But I wouldn`t suggest doing it because we would need to load the tables manually. In the future, we can add some individual behavior that describes what should happen if the hyrise is down.

## Todo 

- [x] Remove `_initialize` in cursor
- [ ] Add timeout to ping


hyrisecockpit/database_manager/cursor.py
```diff
@@ -26,9 +28,19 @@ class PoolCursor:
...
+        """Initialize connection and cursor."""
+        try:
+            self._connection = self.pool.getconn()
+            self._connection.set_session(autocommit=True)
+            self._cur = self._connection.cursor()
+            self.valid = True
+        except (DatabaseError, InterfaceError):
+            self._connection = None
+            self._cur = None
+            self.valid = False
```
heseding: How do we benefit from using an invalid cursor? If the cursor cannot establish a connection, I'd expect this to be raised as an exception.

Alexander-Dubrawski: If we are using an invalid curser, all the components that are using the pool cursor can work normally. If we through an exception (I think this would better but more komplex) every component that uses the pool cursor (worker, backgroundjobs, ...) needs to handle this exception. I decided to implement the exception logic just in the cursor. If you are really unhappy with this approach, what I would understand, we could move the exception handling to the components itself. But we would need to keep in mined that this solution would be more complex and the exception handling would be spread through the components.

hyrisecockpit/database_manager/cursor.py
```diff
@@ -26,9 +28,13 @@ class PoolCursor:
     def __init__(self, connection_pool: pool) -> None:
         """Initialize a PoolCursor."""
         self.pool: pool = connection_pool
-        self.connection = self.pool.getconn()
-        self.connection.set_session(autocommit=True)
-        self.cur = self.connection.cursor()
+        self._initialize()
+
+    def _initialize(self) -> None:
+        """Initialize connection and cursor."""
```
heseding:
```suggestion
```

heseding (approved):

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -79,8 +83,18 @@ def close(self) -> None:
...
         self._update_plugin_log_job.remove()
         self._update_queue_length_job.remove()
+        self._ping_hyrise_job.remove()
         self._scheduler.shutdown()
 
+    def _ping_hyrise(self) -> None:
+        """Check in interval if hyrise is still alive."""
+        try:
+            with PoolCursor(self._connection_pool) as cur:
+                cur.execute("SELECT 1;", None)
```
caterinamandel98, line 120: What if the database is blocked - so we aren't able to execute queries - but still alive?

Alexander-Dubrawski: When the database is blocked we can still execute this query. Here we are not checking the blocked flag. The `SELECT 1;` query shouldn't be a problem in this situation because we are not accessing some meta-information or tables in the hyrise database.

heseding: I believe @caterinamandel98 is not referring to the _blocked_ state we use internally for loading data, but to a _blocked_ state in which the database cannot keep up with the queries.

Alexander-Dubrawski: Ok, then I would suggest using a timeout.

PeterTsayun: Is that possible to retrieve the pending/execution time of `SELECT 1;`-Request in the `ping`-method? In that way, we could return it in `/status` and show in the frontend, how long does it take to reach the database. When a certain threshold is reached, a warning could be shown.

PR #463, opened 2020-03-30
B/use continuous queries

### LINKED ISSUE #466: Requesting not existing Influx database
**Please describe the bug:**  
Sometimes, when adding a new database to `DatabaseManager`, the following error occurs (see traceback below). The reason could be, that when adding a new database, the corresponding Influx database is created when the first job of `BackgroundScheduler` has been executed (`__enter__` in `Cursor`), which can take some time. During this time period, the database is in `active_databases`, but still hasn't the corresponding database in Influx, which leads to the error.

1. Start Backend
2. Add a database
3. Request `/throughput`

`DatabaseManager` should still return correct metric values. One possible solution could be to create an Influx database directly during the initialization of the database object in `DatabaseManager`. 

Further solution ideas are welcome.

**Screenshot(s):**  
```

Traceback (most recent call last):
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/flask_restx/api.py", line 375, in wrapper
    resp = resource(*args, **kwargs)
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/flask_restx/resource.py", line 44, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/home/Petr.Tsayun/Cockpit/hyrisecockpit/backend/app.py", line 576, in get
    result = storage_connection.query(
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/influxdb/client.py", line 458, in query
    results = [
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/influxdb/client.py", line 459, in <listcomp>
    ResultSet(result, raise_errors=raise_errors)
  File "/home/Petr.Tsayun/.local/share/virtualenvs/Cockpit-j4erCpM8/lib/python3.8/site-packages/influxdb/resultset.py", line 25, in __init__
    raise InfluxDBClientError(self.error)
influxdb.exceptions.InfluxDBClientError: database not found: york

```

**Affected Component(s):**  
`App`, `Database`

**Additional context:**

PeterTsayun:
# Resolves #466

In order to provide historical data, we need a simple and consistent way of retrieving throughput and latency information. Continuous queries, provided by InfluxDB, are suitable for that, as they do downsampling of raw data automatically. 
 
Now we are using continuous queries for throughput and latency calculation. On the start of every whole second, Influx runs continuous queries on the last 5 seconds of data and calculates and updates throughput and latency values grouped by every whole second.

Example (Influx shell):
```
> select * from throughput;
name: throughput
time                throughput
----                ----------
1585581950000000000 208
1585581951000000000 261
1585581952000000000 275
1585581953000000000 283
1585581954000000000 281
1585581955000000000 294
1585581956000000000 302
1585581957000000000 293
```

Moreover, by adding a new database to `DatabaseManager`, the corresponding Influx database is now emptied (dropped and created one more time).

**Affected Component(s):**  
`InfluxDB`, `App`

**Additional context:**


hyrisecockpit/database_manager/cursor.py
```diff
@@ -99,6 +99,26 @@ def __write_point(self, point: Point) -> None:
...
+    def create_continuous_query(
+        self,
+        query_name: str,
+        query: str,
+        database_name: str,
+        resample_options: Optional[str] = None,
+    ) -> None:
+        """Create continuous query."""
+        self._connection.create_continuous_query(
+            query_name, query, database_name, resample_options
```
heseding: The `StorageCursor` is specific to one database. We probably shouldn't expose methods to create and drop other databases than the one it was initialized with.

```suggestion
    def create_database(self) -> None:
        """Create database with a given name."""
        self._connection.create_database(self._database)

    def drop_database(self) -> None:
        """Drop database with a given name."""
        self._connection.drop_database(self._database)

    def create_continuous_query(
        self,
        query_name: str,
        query: str,
        resample_options: Optional[str] = None,
    ) -> None:
        """Create continuous query."""
        self._connection.create_continuous_query(
            query_name, query, self._database, resample_options
```

hyrisecockpit/backend/app.py
```diff
@@ -563,9 +563,8 @@ class Throughput(Resource):
     @monitor.doc(model=[model_throughput])
     def get(self) -> Union[int, Response]:
         """Return throughput information from the stored queries."""
-        currentts = time_ns()
-        startts = currentts - 2_000_000_000
-        endts = currentts - 1_000_000_000
+        offset_seconds = 3
+        offsetts = (int(time()) - offset_seconds) * 1_000_000_000
```
heseding, line 567: How about we keep using `time_ns` as it's more precise?

heseding (commented):
The routes `/monitor/latency` and `/monitor/throughput` have been replaced by `/monitor/detailed_query_information` and are no longer used by the frontend.

hyrisecockpit/database_manager/cursor.py
PeterTsayun: Thanks for the suggestion, we could do so

heseding (approved):

PeterTsayun:
> The routes `/monitor/latency` and `/monitor/throughput` have been replaced by `/monitor/detailed_query_information` and are no longer used by the frontend.

Thanks for the information, I didn't know that.

As far as I know, `/monitor/detailed_query_information` returns throughput and latency values grouped by the query type. The total throughput can be calculated as the sum of throughputs of the query types. But how is calculated the total latency? In that way, the latency of every query type should be multiplied with the corresponding throughput value and then divided by the total throughput. The sum of those values should be the average total latency of the last second, which sounds a little bit complicated.

Moreover, `/monitor/detailed_query_information` is shown in the table of the most expensive queries in frontend. If I am not mistaken, last week, our chair wanted to see information about most expensive queries not in the last second, but in the last 5 seconds. After this adjustment,   `/monitor/detailed_query_information` would be no more suitable for calculation of the total throughput/latency information of the last second. Do I see it correctly?

hyrisecockpit/backend/app.py
PeterTsayun: Good question.

Currently, we have some `BackgroundScheduler` jobs, which log information to the InfluxDB. Because they run in parallel, they have different timestamps. So, using precise `time_ns` could lead to the situation, when different metrics will be shown with the different offsets in the graphs in the frontend. 

I looked for ways of solving that in InfluxDB. Let's suppose, we have timepoints `t` and `t+1`. Continuous queries calculate throughput/latency values in the way, that throughput measurement at the timepoint `t` corresponds not to the throughput at exactly `t`, but to the number of executed queries in the time interval between `t` and `t+1`. So, if we would use `int(time())`, for `queue_length`, it would be interpreted as the "average" queue length in the time interval between `t`, `t+1`. So, in that way, all metrics would have the same interpretation and the same timestamps in all graphs.
  
But of course, it is open for a discussion, I don't want to force that approach :blush:

heseding: Makes sense, thanks for the explanation. I suggest we put it on the agenda for the next backend meeting.

PR #464, opened 2020-03-30
B/Log queue length values to InfluxDB

PeterTsayun:
# Resolves #number

In order to provide historical data, we need to log queue length to InfluxDB.

Now, `BackgroundScheduler` logs queue length to InfluxDB every second.

**Affected Component(s):**  
`Database`, `BackgroundScheduler`

**Additional context:**  
The logged queue length data is not used in `App` yet, as it would change the interface of the endpoint. The endpoint will be changed in the historical data PR anyway.

### DIFF
--- a/hyrisecockpit/database_manager/background_scheduler.py
+++ b/hyrisecockpit/database_manager/background_scheduler.py
@@ -83,18 +83,17 @@ class BackgroundJobManager(object):
 
     def _update_queue_length(self) -> None:
         queue_length: int = self._worker_pool.get_queue_length()
-        if queue_length > 0:
-            time_stamp: int = int(time()) * 1_000_000_000
-            with StorageCursor(
-                self._storage_host,
-                self._storage_port,
-                self._storage_user,
-                self._storage_password,
-                self._database_id,
-            ) as log:
-                log.log_meta_information(
-                    "queue_length", {"queue_length": queue_length}, time_stamp,
-                )
+        time_stamp: int = int(time()) * 1_000_000_000
+        with StorageCursor(
+            self._storage_host,
+            self._storage_port,
+            self._storage_user,
+            self._storage_password,
+            self._database_id,
+        ) as log:
+            log.log_meta_information(
+                "queue_length", {"queue_length": queue_length}, time_stamp,
+            )
 
     def _update_krueger_data(self) -> None:
         time_stamp = time_ns()
--- a/hyrisecockpit/database_manager/background_scheduler.py
+++ b/hyrisecockpit/database_manager/background_scheduler.py
@@ -3,7 +3,7 @@
 from copy import deepcopy
 from json import dumps
 from multiprocessing import Process, Value
-from time import time_ns
+from time import time, time_ns
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from apscheduler.schedulers.background import BackgroundScheduler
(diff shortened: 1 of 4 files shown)


tests/database_manager/test_background_scheduler.py
```diff
@@ -58,6 +58,13 @@ def get_mocked_storage_cursor(*args):
     return mocked_storage_cursor_constructor
 
 
+def get_mocked_worker_pool(*args) -> MagicMock:
+    """Return fake storage cursor."""
```
heseding:
```suggestion
    """Return fake worker pool."""
```

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -72,8 +78,24 @@ def close(self) -> None:
         self._update_chunks_data_job.remove()
         self._update_storage_data_job.remove()
         self._update_plugin_log_job.remove()
+        self._update_queue_length_job.remove()
         self._scheduler.shutdown()
 
+    def _update_queue_length(self) -> None:
+        queue_length: int = self._worker_pool.get_queue_length()
+        if queue_length > 0:
+            time_stamp: int = int(time()) * 1_000_000_000
```
heseding: How about using `time_ns` as it's more precise?

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -72,8 +78,24 @@ def close(self) -> None:
         self._update_chunks_data_job.remove()
         self._update_storage_data_job.remove()
         self._update_plugin_log_job.remove()
+        self._update_queue_length_job.remove()
         self._scheduler.shutdown()
 
+    def _update_queue_length(self) -> None:
+        queue_length: int = self._worker_pool.get_queue_length()
+        if queue_length > 0:
```
heseding: Why should we not log an empty queue? Also, can the queue have a negative length?

hyrisecockpit/database_manager/background_scheduler.py
PeterTsayun: Good question.

Currently, we have some `BackgroundScheduler` jobs, which log information to the InfluxDB. Because they run in parallel, they have different timestamps. So, using precise `time_ns` could lead to the situation, when different metrics will be shown with the different offsets in the graphs in the frontend. 

I looked for ways of solving that in InfluxDB. Let's suppose, we have timepoints `t` and `t+1`. Continuous queries calculate throughput/latency values in the way, that throughput measurement at the timepoint `t` corresponds not to the throughput at exactly `t`, but to the number of executed queries in the time interval between `t` and `t+1`. So, if we would use `int(time())`, for `queue_length`, it would be interpreted as the "average" queue length in the time interval between `t`, `t+1`. So, in that way, all metrics would have the same interpretation and the same timestamps in all graphs.
  
But of course, it is open for a discussion, I don't want to force that approach :blush:

hyrisecockpit/database_manager/background_scheduler.py
PeterTsayun: Queue length returns the numbers of the items in the queue and is thus always positive. I wanted to save memory usage in Influx and interpret missing entry for queue length at the specific timepoint as a queue length of the size 0. But, of course, we could also simply log it.

heseding: I'm a fan of logging a queue length of 0 as well, how about we put it up for discussion during our next backend meeting?

heseding (approved):

PR #467, opened 2020-03-31
Add CRUD for workloads in WorkloadGenerator

### LINKED ISSUE #432: Show a running workload indicator
Despite the throughput spike and treemap, there is no indicator whether a workload is running.
In addition it is not visible what workload is currently running.

The slider / radio button in the workload menu should indicate what workload is currently running.

**Affected Component(s):**  
`Frontend`, `Backend`

reported by @Bensk1

### LINKED ISSUE #434: Backend endpoint for query distribution
Queries are chosen with the same probability. However, this probability should be set by the user.

The existing Workload endpoint should expose frequency settings for each query of the workload.

An additional endpoint, but that would clutter the long list of endpoints.

**Affected Component(s):**  
`CockpitBackend`, `WorkloadGenerator`

**Additional context:**  
Like that:
image

### LINKED ISSUE #544: Workload api inconsistent
**Please describe the bug:**  
Now, the workload API is in an inconsistent state, as the endpoints of the API don't suit the `WorkloadGenerator`-interface.

1. Start backend
2. Add database
3. Start workload

The user should be able to start the workload execution.

**Screenshot(s):**  
https://github.com/hyrise/Cockpit/blob/2a60cd55237c51edb652cf6851f5c570953653fb/hyrisecockpit/api/app/workload/service.py#L58-L62
In the current implementation, `WorkloadGenerator` returns a response with an empty body, causing the error at line 61.
https://github.com/hyrise/Cockpit/blob/2a60cd55237c51edb652cf6851f5c570953653fb/hyrisecockpit/api/app/workload/service.py#L15-L19
Moreover, the generator URL is hardcoded and independent from `.env`. 

**Affected Component(s):**  
`Api`

**Additional context:**

heseding:
# Resolves #432, resolves #434, resolves #544

The `WorkloadGenerator` is not able to run multiple workloads at once.
Additionally, queries are chosen sequentially and with equal weights as the only option.
Finally, the new backend namespace `/workload` was not used.

- 🏃 Multiple workloads can be run at once (e.g. `job` and `tpch_1`)
- `WorkloadGenerator` can, in addition to starting and stopping workloads,
  - list all running workloads (such that the frontend can show a running workload indicator)
  - get one by `folder_name` (also showing weights)
  - update a workload by `folder_name` (with weights)
- The backend namespace `/workload` now replaces the old `/control/workload` endpoint

**Affected Component(s):**  
`WorkloadGenerator`, `API`

**Additional context:**  

Weights can be set for each query of a workload (initially, all weights are set to 100), such as:
```python
# Example weights for each query type of an example workload
{
  "01": 0,
  "02": 5,
  "03": 999,
}
```

The environment setting `WORKLOAD_LOCATION` has been removed since the location is tracked in the repo anyways. The workloads are now located in `workloads/` instead of `workloads/workload_queries/` since there's no reason for the nested directory anymore.

**Bonus:**  
- [ ] Test `WorkloadGenerator`
  - [x] Test `_call_start_workload`
  - [x] Test `_call_stop_workload`
  - [ ] Test `_call_get_all_workloads`
  - [ ] Test `_call_get_workload`
  - [ ] Test `_call_update_workload`
- [ ] Test `WorkloadReader`
- [ ] Test `WorkloadIdController.test_updates_the_correct_workload` (it's skipped)
- [x] Test `WorkloadService` ~(it's skipped)~

@schTi


hyrisecockpit/workload_generator/generator.py
```diff
@@ -80,49 +71,79 @@ def _init_server(self) -> None:
...
+        return response
 
     def _call_start_workload(self, body: Body) -> Response:
+        folder_name: str = body["folder_name"]
         frequency: int = body["frequency"]
-        workload_type: str = body["folder_name"]
+        if folder_name not in self._workloads:
+            queries = WorkloadReader.get(folder_name)
+            if queries is not None:
+                self._workloads[folder_name] = Workload(frequency, queries)
```
PeterTsayun, line 89: What if the user would like to set weights directly when starting a workload?

hyrisecockpit/workload_generator/generator.py
```diff
@@ -80,49 +71,79 @@ def _init_server(self) -> None:
...
-            queries = workload.generate_workload(self._frequency)
+        for folder_name, workload in self._workloads.items():
             response = get_response(200)
-            response["body"] = {"querylist": queries}
-            self._publish_data(response)
+            response["body"]["querylist"] = [
+                (query.query, query.args, folder_name, query.query_type)
+                for query in workload.get()
+            ]
+            self._pub_socket.send_json(response)
```
PeterTsayun: Would it be possible to send all workload queries at once rather than send them once for each workload type?

heseding: Sure.

hyrisecockpit/workload_generator/generator.py
heseding: Initially, I wanted to allow just that, but this would mean the user (frontend) should know the queries beforehand, which is not the case (there won't be any data in the frontend about which queries belong to which workload).
We provide this with the `get_by_id`, which returns all weights (thus all queries belonging to the workload) and let the user update the weights once he knows which queries exist.

We could provide the possibility to set the weights initially, but it would introduce further complexity and the need for data mirroring our data in the frontend, which I believe to be unnecessary.

PeterTsayun: Okay, I get the point. Just as an idea:

In order to start a workload, the frontend needs to know, which workloads are supported. If I am not mistaken, this information is provided by `GET /workload` endpoint (or `GET /data`), where we return a list of available workloads, something like `[tpch_0.1, job, ...]`. What, if we would provide a little bit more precise information, like together with the query names, like:
```
[
    {
        "benchmark": "tpch_0.1",
        "query_types": [
            "1",
            "2a",
            "2b",
            ...
        ]
    }
]
```

In that way, the frontend could ask it once when starting and then have complete information about all workload equalizers to be shown.

But, of course, we could simply ask, what is the use case.

heseding: That is definitely a promising solution, but as you mentioned: what is the use case?😅
How about this: We continue this way with the weights via `GET` and `PUT`, and if that proves cumbersome, we'll add it to `POST` as well?

hyrisecockpit/api/app/workload/controller.py
```diff
@@ -23,49 +23,40 @@ def get(self) -> List[Workload]:
...
-        else:
-            return Response(status=404)
-
-    @api.response(404, "A Workload with the given ID doesn't exist.")
-    @accepts(schema=WorkloadSchema, api=api)
-    @responds(schema=WorkloadSchema, api=api)
-    def put(self, workload_id: str) -> Union[Workload, Response]:
+        return (
+            Response(status=200)
+            if WorkloadService.delete_by_id(folder_name)
```
caterinamandel98, line 52: Shouldn't we rename these if we don't delete/update/.. by id anymore?

heseding: `folder_name` is still the primary key, and I believe `delete_by_id` to convey that the primary key is used (even across different primary key names).
Given that other components will also have a primary key used for CRUD operations, renaming all of them to include the specific name of the primary key could hurt readability (e.g. `delete_by_folder_name`, `delete_by_database_id`, ...).

hyrisecockpit/workload_generator/generator.py
PeterTsayun: I've just talked to the frontend team, and they would prefer to have full query information of the workloads at the beginning when the site is loaded, as mentioned above.

The use case is, that in the current solution in order to adjust a workload, it should be started first if I see it correctly. So, the actual workload equalizer can only be built after the start of the workload execution, which sounds a little bit unclean in my opinion.

hyrisecockpit/api/app/workload/controller.py
```diff
@@ -23,49 +23,40 @@ def get(self) -> List[Workload]:
...
     def post(self) -> Union[Workload, Response]:
         """Create a Workload."""
-        workload = WorkloadService.create(request.parsed_obj)
-        if workload is None:
-            return Response(status=409)
-        else:
-            return workload
+        interface: WorkloadInterface = request.parsed_obj
+        workload = WorkloadService.create(interface)
+        return Response(status=409) if workload is None else workload
```
Alexander-Dubrawski, line 33: Mabey we can just simply return the error code as an integer from the service. Then we can simply return it in the response and we don't need to check if the answer from the service is None.

hyrisecockpit/api/app/workload/service.py
```diff
@@ -52,69 +44,81 @@ def create(cls, interface: WorkloadInterface) -> Optional[Workload]:
...
+            response = cls._send_message_to_gen(
+                Request(header=Header(message="start workload"), body=dict(interface)),
+            )
+            return (
+                None
+                if response["header"]["status"] == 409
+                else Workload(**response["body"]["workload"])
+            )
+        else:
+            return None
```
Alexander-Dubrawski, line 58: Like mentioned before we could return the error code as an integer directly to the controller 🙃.

hyrisecockpit/workload_generator/generator.py
Alexander-Dubrawski: I think we shouldn't but too much complexity in the endpoints. In my opinion, it is just to complex if we set the equalizer, start the workload and start the workers. We should provide three endpoints for these three logical steps. In the end, we will have much more freedom and clarity, as well as flexibility. #505 implements an endpoint to start the worker separately. So we will be able to start the workload, set the equalizer if we want and then start the worker.  I think that is a clean approach.

hyrisecockpit/api/app/workload/controller.py
heseding: That's definitely possible and might be more flexible in terms of adding new error events, however, we have not added error codes previously. On the other hand, checking for specific responses (i.e. `None`) and explicitly returning a predefined `Response` is easier to grasp and increases readability. Would you be okay with this approach?

hyrisecockpit/workload_generator/generator.py
heseding: I'm with @Alexander-Dubrawski on this.

hyrisecockpit/api/app/workload/controller.py
Alexander-Dubrawski: It's just some stylistic choice, so I'm fine with it 🤓.

Alexander-Dubrawski (dismissed):

Alexander-Dubrawski (approved):

caterinamandel98 (approved):

PR #473, opened 2020-04-01
f/471-adapt-selection

### LINKED ISSUE #471: adapt selection
Currently the selection is not really cool. Thats why it has to be improved like this:

image

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #471

The selection handling wasnt really good. We want to improve this.

This PR adds a selection menu, which selected databases and metrics separated for every page.

**Affected Component(s):**  
`Frontend`

**Additional Context**
Thats how it looks like:

image


hyrisecockpit/frontend/src/components/details/MetricChip.vue
```diff
@@ -0,0 +1,65 @@
...
+      left
+      @click="handleUnSelect(metric)"
+      >mdi-eye</v-icon
+    >
+    <v-icon
+      v-if="!selected"
+      id="remove-select-metric-button"
+      left
+      @click="handleSelect(metric)"
+      >mdi-eye-off</v-icon
```
MaxSchneider1337, line 15: could you tell me the difference in implementation for both chips forms?
Could we just use a chip-component  and only give the db chip an individual color?

hyrisecockpit/frontend/src/components/selection/DatabaseSelection.vue
```diff
@@ -0,0 +1,95 @@
+<template>
```
MaxSchneider1337, line 1: same here, why not just use a selection component and compose it with databases or metrics?

MaxSchneider1337: AAAh ok, I think I understood your architecture now, could you just describe it shortly, for me to validate my assumption

hyrisecockpit/frontend/src/components/details/MetricChip.vue
schTi: I thought about this too, but there are some differences between database and metric chips. The database chip contains multiple icons, and has some optional closing handling (for the database removal). In my opinion the biggest similarity is the chip component, which is already an own vueitfy component. It would be an overhead  and not really clean to handle those edge cases, as I think.

hyrisecockpit/frontend/src/components/selection/DatabaseSelection.vue
schTi: Here are some differences, too. Maybe some functions can be shared, but Im not sure if this is necessary yet. The biggest difference is in available item handling. For metrics it is always clear, which metrics are available on which page. But the available databases are changing over time, and it need to be waited for their first fetch in general. In contrast the available metrics need handling for page changes. But I see your point, maybe we should consider to refactor this at some point. When thinking about selecting queries later, too, it will give us more flexibility now.

hyrisecockpit/frontend/src/views/AppDrawer.vue
```diff
@@ -1,41 +1,25 @@
...
-  <v-navigation-drawer v-model="drawer" app fixed width="250">
+  <v-navigation-drawer app fixed width="250">
     <v-list>
       <v-list-item two-line>
         <v-list-item-avatar tile>
           <img src="../../src/assets/images/hyrise_logo.png" />
         </v-list-item-avatar>
         <v-list-item-content>
-          <v-list-item-title>Cockpit</v-list-item-title>
+          <v-list-item-title><b>Cockpit</b></v-list-item-title>
```
monasobh, line 9: Its just a small comment. I think we will have to change the font overall a little, to make it all look better.

hyrisecockpit/frontend/src/views/AppDrawer.vue
```diff
@@ -76,9 +60,13 @@
...
         <template v-slot:activator="{ on: menu }">
-          <v-list-item id="database-list-button" v-on="{ ...menu }">
+          <v-list-item
+            id="database-list-button"
+            v-on="{ ...menu }"
+            @click="$emit('closeSelection')"
+          >
             <v-list-item-icon>
-              <v-icon>mdi-database-plus</v-icon>
+              <v-icon>mdi-database</v-icon>
```
monasobh, line 69: Ah, did you delete the plus because we are now also able to remove a database?

monasobh (commented):
cool! I really like the new selection button. 👍

hyrisecockpit/frontend/src/views/AppDrawer.vue
schTi: Yeah I agree with that

hyrisecockpit/frontend/src/views/AppDrawer.vue
MaxSchneider1337: I think the plus should not be there, because this shows the database list, which is there to show all available databases. In normal frontend-design you use these lists to also add and remove Items. So to only show that there is a db-list, the db-icon seems to be just right

schTi: Yes exactly. And it fits more with the Setttings for Plugins and Workload

MaxSchneider1337 (approved):
n1

PR #474, opened 2020-04-02
U/add multi process wrk plug in

Alexander-Dubrawski:
This PR implements the `multiProcessWrk` Plug-in. We can now stress the endpoints at the same time so we can measure our capability to process parallel requests.


Bouncner:
Not the biggest fan of abbreviations, but some are just standard (MP for multi processing). But the letters you safe by using `multiProcessWrk` over `multiProcessWorker` isn't really worth, is it?

PeterTsayun (approved):

Alexander-Dubrawski:
> Not the biggest fan of abbreviations, but some are just standard (MP for multi processing). But the letters you safe by using multiProcessWrk over multiProcessWorker isn't really worth, is it?

This PR has nothing to do with the Backend 🙃. We have a utility tool to measure some performance aspects of your backend. 

https://github.com/hyrise/Cockpit/tree/dev/utils/endpoint_benchmark

For this utility tool, we can write some Plugins and use them. `multiProcessWrk` is a new plugin that uses the `wrk` benchmark tool on every endpoint at the same time.

https://github.com/wg/wrk

We have some more Plugins as well, like:

* sequenziall wrk plugin 
* A Plugin that displays the Replay of the endpoints

Bouncner:
> > Not the biggest fan of abbreviations, but some are just standard (MP for multi processing). But the letters you safe by using multiProcessWrk over multiProcessWorker isn't really worth, is it?
> 
> This PR has nothing to do with the Backend 🙃. We have a utility tool to measure some performance aspects of your backend.
> 
> https://github.com/hyrise/Cockpit/tree/dev/utils/endpoint_benchmark
> 
> For this utility tool, we can write some Plugins and use them. `multiProcessWrk` is a new plugin that uses the `wrk` benchmark tool on every endpoint at the same time.
> 
> https://github.com/wg/wrk
> 
> We have some more Plugins as well, like:
> 
> * sequenziall wrk plugin
> * A Plugin that displays the Replay of the endpoints

Then please open an issue for `wrk` that they are stupid.

PR #475, opened 2020-04-02
B/add multithreading to thread safe endpoints

Alexander-Dubrawski:
This PR enables multithreaded Endpoints in Flask 🎉. There are still some endpoints that are not threadsafe. The Threads that are processing these kinds of endpoints need to wait for a Lock. 
All requests to the `manager` are not thread-safe. 

https://docs.python.org/3/library/multiprocessing.html#multiprocessing.RLock

Because in most of the endpoints we need to access the information which databases are registered in the manager, we are saving this information in the app itself as well.

**Affected Component(s):**  
`App`


hyrisecockpit/backend/app.py
```diff
@@ -541,19 +544,27 @@
...
+
+def _get_active_databases() -> List[str]:
     """Get a list of active databases."""
-    response: Response = _send_message(
-        db_manager_socket, {"header": {"message": "get databases"}, "body": {}}
-    )
-    validate(instance=response["body"], schema=get_databases_response_schema)
-    return [database["id"] for database in response["body"]["databases"]]
+    global active_databases
+    return active_databases
```
heseding, line 567: Are these methods necessary? We could just use `.append` and `.remove` directly on `active_databases`, and use the `active_databases` list directly without `_get_active_databases`.

hyrisecockpit/backend/app.py
```diff
@@ -568,7 +579,7 @@ def get(self) -> Union[int, Response]:
 
         throughput: Dict[str, int] = {}
         try:
-            active_databases = _active_databases()
+            active_databases = _get_active_databases()
```
heseding: There's no need to catch a `ValidationError` since we're not using `send_message` anymore.

hyrisecockpit/backend/app.py
Alexander-Dubrawski: I think it is better to keep the logic capsuled for now.

PeterTsayun (approved):

hyrisecockpit/backend/app.py
```diff
@@ -630,10 +635,7 @@ def get(self) -> Union[int, List[Dict[str, Any]]]:
         currentts = time_ns()
         startts = currentts - 2_000_000_000
         endts = currentts - 1_000_000_000
-        try:
-            active_databases = _active_databases()
-        except ValidationError:
-            return 500
+        active_databases = _get_active_databases()
```
PeterTsayun, line 638: Could it actually happen, that one process would read `active_databases` and before it executes the query, `active_databases` would be no more valid, because some other thread has changed it in the meantime?

hyrisecockpit/backend/app.py
heseding: Why?

hyrisecockpit/backend/app.py
Alexander-Dubrawski: This is very unlikely. But in an event where this would happen, we would just get this information one more time in the endpoint for the database (which is not active anymore). This is not critical because one second later at the next request we would see that this database is not active anymore. In all critical operations where we need to communicate with the `manager`, we are not accessing the active database list.

hyrisecockpit/backend/app.py
Alexander-Dubrawski: We have the logic that handles the management of the active database list. I think this is one central unit in the code and I like to control the access to it via *getter* and *setter*. It shouldn't be the task of the endpoint classes to directly change the stats of the active databases list. I think that way we just have a better separation in the code.

PR #476, opened 2020-04-02
F/456-database-colors

### LINKED ISSUE #456: Database Colors
**Please describe the bug:**  
The database color changes.

The color should be consistent for a database, even on reload

**Affected Component(s):**  
`Frontend`

kathwill:
# Resolves #456 

The database color changes when starting the frontend, reloading or adding a database.

There are now consistent colors for the database instances.

**Affected Component(s):**  
`Frontend`


hyrisecockpit/frontend/src/meta/colors.ts
```diff
@@ -33,3 +33,10 @@ export const colorDefinition: Record<Colors, Hex> = {
   darkred: "#e31a1c",
   darkviolet: "#6a3d9a"
 };
+
+export const databaseColorDefinition: Record<string, any> = {
```
schTi: Im not sure if this is what we really wanted here. You just hardcoded some IDs, but when using a different ID the ID will not be persistent. But I am not sure if we really want to have persistent coloring. What do you think @MaxSchneider1337?

schTi (changes requested):

hyrisecockpit/frontend/src/meta/colors.ts
kathwill: I thought we wanted persistent coloring for our default instances. Otherwise we would  need to know all the IDs that were used before

schTi: No we could use a hash function which will always assign an ID to the same color

MaxSchneider1337: Persistent coloring was requested so we definitely need it. I would also suggest using an hash function to determine the color for a database

hyrisecockpit/frontend/src/services/databaseService.ts
```diff
@@ -31,11 +30,14 @@ export function useDatabaseService(): DatabaseService {
   }
 
   function getDatabaseColor(databaseID: string): string {
-    let color = databaseColorDefinition[databaseID];
-    if (color === undefined) {
-      color = Object.values(colorDefinition)[usedColors];
-      usedColors += 1;
-    }
+    const hashedDatabaseID = require("crypto")
```
schTi: The npm crypto package is deprecated, please move to crypto-js instead.

schTi: As I think you need to include the new package to the dev dependencies before using it.

schTi (changes requested):

hyrisecockpit/frontend/src/services/databaseService.ts
```diff
@@ -31,9 +29,15 @@ export function useDatabaseService(): DatabaseService {
...
-    usedColors += 1;
+  function getDatabaseColor(databaseID: string): string {
+    const crypto = require("crypto-js");
+    const hashedDatabaseID = crypto
+      .SHA1(databaseID)
+      .toString(crypto.enc.Hex)
+      .substring(0, 8);
+    const index =
+      parseInt(hashedDatabaseID, 16) % Object.keys(colorDefinition).length;
+    let color = Object.values(colorDefinition)[index];
```
Bouncner, line 45: Sorry for these direkt comments. Github is a mess.

I asked if makes sense to use the host name, but I am rather asking for hostname+port.

kathwill: We are using the ID of the database here. You can specify it when adding a database (the default is the hostname, e.g. "vm-york", ...)

Bouncner: What happens if we have two instances on the same node?
How is this database ID determined?

schTi: It is not possible to add two instances with the same ID. You have to change the ID manually before adding the new instance.

Bouncner: Do you have screenshot of the add-instance-screen at hand?

schTi: image

Bouncner: The port is hidden in advanced?
And you default the database id to the host name?

hyrisecockpit/frontend/src/services/databaseService.ts
```diff
@@ -31,9 +29,18 @@ export function useDatabaseService(): DatabaseService {
...
+    Object.values(databaseID).forEach((symbol: any) => {
+      hashedDatabaseID =
+        (hashedDatabaseID << 5) -
+        hashedDatabaseID +
+        databaseID.charCodeAt(symbol);
+      hashedDatabaseID = hashedDatabaseID & hashedDatabaseID;
+    });
+    const index =
+      Math.abs(hashedDatabaseID) % Object.keys(colorDefinition).length;
+    let color = Object.values(colorDefinition)[index];
```
Bouncner, line 45: How do you determine that two instances do not get the same color index?

hyrisecockpit/frontend/src/services/databaseService.ts
```diff
@@ -31,9 +29,18 @@ export function useDatabaseService(): DatabaseService {
     return databases;
   }
 
-  function getDatabaseColor(): string {
-    const color = colors[usedColors];
-    usedColors += 1;
+  function getDatabaseColor(databaseID: string): string {
+    let hashedDatabaseID = 0;
+    Object.values(databaseID).forEach((symbol: any) => {
+      hashedDatabaseID =
```
Bouncner, line 37: Explanation?

hyrisecockpit/frontend/src/services/databaseService.ts
schTi: Yes exactly. When typing the host field, it will automatically add this input to the ID field., until you manually change the ID

Bouncner: Ok, got it.

schTi (approved):

Bouncner:
I am surprised that this was merged even though there are still a lot of unresolved issues.

hyrisecockpit/frontend/src/services/databaseService.ts
MaxSchneider1337: @kathwill I already ran into that problem I think.

MaxSchneider1337:
@kathwill please resolve all conservations and answer all questions before merging the pr

hyrisecockpit/frontend/src/services/databaseService.ts
Bouncner: @MaxSchneider1337 : can you please open an issue for that? @kathwill: can you take that issue and make a recommendation how to approach this problem? It's not trivial. We should discuss this.

PR #482, opened 2020-04-03
f/479-change-time-interval

### LINKED ISSUE #479: change time interval
We dont need a specific time interval shown on top of every chart type yet. It is enough to just show of which time interval type they are. (e.g. line charts: last 1 second ...)

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #479

It is not necessary to show the current time stamps of the data above the charts. Instead it is necessary to easily determine, what the displayed data means. Therefore we need to know which interval will be aggregated. 

This PR will replace the current time stamps with this information. For storage data we will instead show the time stamp of the last snapshot.

**Affected Component(s):**  
`Frontend`

image

### DIFF
--- a/hyrisecockpit/frontend/src/components/charts/Barchart.vue
+++ b/hyrisecockpit/frontend/src/components/charts/Barchart.vue
@@ -56,7 +56,7 @@ function useBarChartConfiguration(
         l: 70,
         r: 40,
         b: 70,
-        t: 10,
+        t: 20,
         pad: 0
       },
       paper_bgcolor: "rgba(0,0,0,0)"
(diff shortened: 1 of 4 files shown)


MaxSchneider1337 (approved):
n1

Bouncner:
What is listed for the access counters heat map?

Bouncner:
I am not too certain that "last second" is a valid interval. Does that really tell a new user what is being shown here? I don't think so.

Please put some more love in such things.

schTi:
What do you prefer to show?
Currently it's 'Interval: last 5 seconds' for heatmap

Bouncner:
It's important to note that we are not showing the absolute numbers but the delta for the 5s interval. I'll let you guys come up with an idea. :)

PR #484, opened 2020-04-03
B/provide historical data

### LINKED ISSUE #264: provide historical data for frontend
we want to show historical data in the frontend.
Right now the frontend only requests the current value and stores the data. This causes the data to be deleted on reload and makes it impossible to show historical data. Also the backend should be the source of truth, right now multiple frontend could store different datasets. 

The endpoints for every metric that is shown in a linegraph(throughput, CPU, RAM, latency, queue-length) should return an array of entries, which always include a value and a timestamp. The general response structure should stay the same, so e.g. the throughput endpoint should still return information about all databases, but with the new historical data. 
The request we send will include a time interval e.g. (Date.now()-30, Date.now()) and the endpoints should return the throughput values for every full second in the interval for all databases. 

Maybe we have to talk about the timestamp format, just ask Til or me. 

If you got any alternate ideas, talk to Til or me.

PeterTsayun:
# Resolves #264

This PR adds historical data to `/throughput`, `/latency`, `/queue_length`, `/system` endpoints. 

In the request, the client needs to provide `startts` and `endts` HTTP parameters for a time interval  - 
`[startts, endts)` in UNIX-format:
```
  "startts": 1585762457000000000,
  "endts": 1585762459000000000
```
Example response (`/throughput`):

```
[
  {
    "id": "hyrise-1",
    "throughput": [
      {
        "timestamp": 1585762457000000000,
        "throughput": 273
      },
     {
        "timestamp": 1585762458000000000,
        "throughput": 285
      },
    ]
  }
]
```
**Affected Component(s):**  
`App`

**Additional context:**  
Frontend wished to have timestamps in the UNIX-format (nanoseconds since epoch). To convert Influx time strings to UNIX-timestamps, following code is used:
```
from iso8601 import parse_date

int(parse_date(row["time"]).timestamp() * 1e9)
```
It has the disadvantage of rounding errors, as `datetime.datetime.timestamp` returns timestamp in in the milliseconds precision. So, any other suggestions for converting are appreciated.

Related to #494.


hyrisecockpit/api/app/monitor/app.py
```diff
@@ -382,26 +553,48 @@ def get(self) -> Union[int, List[Dict[str, Any]]]:
...
+        model=[model_latency],
+        params={
+            "startts": "Start of a time interval",
+            "endts": "End of a time interval",
+        },
+    )
+    def get(self) -> Union[int, List]:
+        """Return latency information in a given time range."""
+        rough_startts: int = int(request.args.get("startts"))  # type: ignore
+        rough_endts: int = int(request.args.get("endts"))  # type: ignore
```
Alexander-Dubrawski: Can we rename rough in something more meaningful? Like I see it is the time Intervall for the historical data. So why not something like  `interval_start_time_stamp`?

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -382,26 +553,48 @@ def get(self) -> Union[int, List[Dict[str, Any]]]:
...
+
+        startts_rounded: int = int(rough_startts / 1_000_000_000) * 1_000_000_000
+        endts_rounded: int = int(rough_endts / 1_000_000_000) * 1_000_000_000
+
+        # take nearest whole numbers of seconds
+        startts: int = startts_rounded if rough_startts % 1_000_000_000 == 0 else startts_rounded + 1_000_000_000
+        endts: int = endts_rounded if rough_endts % 1_000_000_000 == 0 else endts_rounded + 1
```
Alexander-Dubrawski: I think a better approach would be to validate the arguments itself. We could use 
https://marshmallow.readthedocs.io/en/stable/
for that. One possibility is to define a `Schema` that checks if the arguments are seconds.

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -382,26 +553,48 @@ def get(self) -> Union[int, List[Dict[str, Any]]]:
...
+            "endts": "End of a time interval",
+        },
+    )
+    def get(self) -> Union[int, List]:
+        """Return latency information in a given time range."""
+        rough_startts: int = int(request.args.get("startts"))  # type: ignore
+        rough_endts: int = int(request.args.get("endts"))  # type: ignore
+
+        startts_rounded: int = int(rough_startts / 1_000_000_000) * 1_000_000_000
+        endts_rounded: int = int(rough_endts / 1_000_000_000) * 1_000_000_000
```
Alexander-Dubrawski: We should just allow seconds as entities. In that case, we don't need to round 🙃 .

hyrisecockpit/api/app/monitor/app.py
PeterTsayun: How about `precise_startts`, as it represents precise timestamp in nanoseconds?

hyrisecockpit/api/app/monitor/app.py
PeterTsayun: We'll refactor `/monitor`-namespace anyway, by providing controllers, services, models, etc from best practices. So I suggest defining schemas in that PR.

hyrisecockpit/api/app/monitor/app.py
PeterTsayun: Yes, but in that case, we are able to provide measurements for any time interval, not only for intervals in whole seconds. The rounding is only used for retrieving information from `/throughput` and `/latency` endpoints, as they are calculated via continuous queries. So, in that special case, points are stored only with seconds-timestamps, which is different from `/queue_length` or `/system`.

Moreover, the frontend wished to use UNIX-timestamps everywhere and stay consistent.

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -472,38 +693,53 @@ def get(self) -> List[Dict[str, Union[str, List]]]:
...
+                        "database_threads": row["database_threads"],
                     },
-                    "database_threads": system_value["database_threads"],
                 }
-            else:
-                system[database] = {}
-        response = get_response(200)
-        response["body"]["system_data"] = system
+                for row in system_data_rows
+            ]
```
heseding, line 740: :fire:

heseding (approved):

PR #494, opened 2020-04-07
f/move-to-historical-data

### LINKED ISSUE #264: provide historical data for frontend

schTi:
# Resolves #264

This PR will add handling for historical data in frontend.

On page change all line charts will now be filled by data of the last 30 seconds. Otherwise they will get new data of the last second, as before. Additionally the implementation allows to handle dynamic selection of historic time ranges, which may be worked on in the future.

**Affected Component(s):**  
`Frontend`


schTi:
TODO:
- fix frontend tests

hyrisecockpit/frontend/src/meta/events.ts
```diff
@@ -36,6 +36,28 @@ export function useMetricEvents(): {
...
+} {
+  function emitPageChangedEvent(metrics: Metric[]): void {
+    eventBus.$emit("PAGE_CHANGED", metrics);
+  }
+  function emitHistoricRangeChangedEvent(
+    metrics: Metric[],
+    start: Date,
+    end: Date
+  ): void {
+    eventBus.$emit("PAGE_CHANGED", { metrics, start, end });
```
MaxSchneider1337: why does this also emit `PAGE_CHANGED`

hyrisecockpit/frontend/src/services/pluginService.ts
```diff
@@ -17,7 +17,7 @@ export function usePluginService(): PluginService {
   } = usePluginTransformationSevice();
 
   getPlugins();
-  setInterval(() => getPluginLogs(), 1000);
+  //setInterval(() => getPluginLogs(), 1000);
```
MaxSchneider1337: you probably do not want to push this

MaxSchneider1337 (commented):
looks good so far

hyrisecockpit/frontend/src/meta/events.ts
schTi: because copy pasta

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -25,15 +25,15 @@ export function useDatabaseController(): DatabaseController {
 
   eventBus.$on("STORAGE_DATA_CHANGED", (data: any) => {
     if (!allDatabasesExist(Object.keys(data))) {
-      updateDatabases();
+      //updateDatabases();
```
MaxSchneider1337, line 28: why is this a comment?

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -25,15 +25,15 @@ export function useDatabaseController(): DatabaseController {
...
     } else {
       updateDatabaseStorageInformation(data);
     }
   });
 
   eventBus.$on("CPU_DATA_CHANGED", (data: any) => {
-    if (!allDatabasesExist(Object.keys(data))) {
-      updateDatabases();
+    if (!allDatabasesExist(data.map((entry: any) => entry.id))) {
+      //updateDatabases();
```
MaxSchneider1337, line 36: see above

MaxSchneider1337 (approved):
just some questions, good changes over all

hyrisecockpit/frontend/src/controller/databaseController.ts
schTi: Just because this caused updating database loops and it triggered me all the time. Will be removed with Database Colors PR anyway

PR #513, opened 2020-04-09
limit fetched plugin log messages

### LINKED ISSUE #490: limit number of fetched plugin log messages
**Please describe the bug and the desired solution:**  
At the moment we collect all plugin log messages from every hyrise instance and only write the new ones in influx. As we don't need the messages that we already collected in the former request, we can just limit our request to new log messages.

**Affected Component(s):**  
`BackgroundManager`

caterinamandel98:
# Resolves #490 

At the moment we fetch all plugin log messages stored in the hyrise instance every second and always overwrite the ones we've already saved in the influx. As we don't need to fetch already stored messages, we limit the newly fetched messages to the ones that have been added in the last second.

**Affected Component(s):**  
`BackgroundJobManager`


Alexander-Dubrawski (approved):

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -241,7 +241,12 @@ def _update_chunks_data(self) -> None:
 
     def _update_plugin_log(self) -> None:
         """Update plugin log."""
-        log_df = self._sql_to_data_frame("SELECT * FROM meta_log;")
+        currentts = time_ns()
+        startts = currentts - 1_000_000_000
+
+        log_df = self._sql_to_data_frame(
+            f"SELECT * FROM meta_log WHERE 'timestamp' >= {startts} AND 'timestamp' < {currentts};",
```
heseding: How about using `bind_params` instead? https://en.wikipedia.org/wiki/SQL_injection

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -241,7 +241,12 @@ def _update_chunks_data(self) -> None:
 
     def _update_plugin_log(self) -> None:
         """Update plugin log."""
-        log_df = self._sql_to_data_frame("SELECT * FROM meta_log;")
+        currentts = time_ns()
+        startts = currentts - 1_000_000_000
```
PeterTsayun: I'm not sure about the internal implementation of Hyrise, but it could potentially be, that the timestamp of the log message is not exactly the same time point, when the message was logged to the `meta_log` table.  So, the actual write to the table could happen a little bit later. Moreover, the `Backgroundscheduler`-jobs run not exactly in 1-second intervals (+- some nano/milliseconds). 

So, what, if we would use the time interval of 2 seconds?

caterinamandel98: Good idea!

PeterTsayun (approved):

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -241,11 +241,14 @@ def _update_chunks_data(self) -> None:
...
         """Update plugin log."""
-        currentts = time_ns()
-        startts = currentts - 2_000_000_000
+        endts = time_ns()
+        startts = endts - 2_000_000_000
 
         log_df = self._sql_to_data_frame(
-            f"SELECT * FROM meta_log WHERE 'timestamp' >= {startts} AND 'timestamp' < {currentts};",
+            (
+                "SELECT * FROM meta_log WHERE 'timestamp' >= startts AND 'timestamp' < entts;"
```
heseding:
```suggestion
                "SELECT * FROM meta_log WHERE 'timestamp' >= startts AND 'timestamp' < endts;"
```

tests/database_manager/test_cursor.py
```diff
@@ -232,7 +232,9 @@ def test_reads_sql_query_pandas(
         pool_cursor._connection = moked_connection
         mocked_read_sql_query_pandas.return_value = fake_df
 
-        results: DataFrame = pool_cursor.read_sql_query("query")
+        results: DataFrame = pool_cursor.read_sql_query("query", None)
 
         assert results.equals(fake_df)
-        mocked_read_sql_query_pandas.assert_called_once_with("query", moked_connection)
+        mocked_read_sql_query_pandas.assert_called_once_with(
+            "query", moked_connection, params=None
```
heseding:
```suggestion
            "query", mocked_connection, params=None
```

heseding (approved):

PR #515, opened 2020-04-14
f/add-dev-server

schTi:
# Resolves #number

Because it is sometimes nasty to develop UI features, when backend is not stable, we need a reliable setup.

This PR adds a basic node server, which just reuses the mocked backend, which is already used for frontend test. To start this server:
- `cd Cockpit/hyrisecockpit/frontend/devServer`
- `npm install`
- `npm start`
--> the server will always listening on localhost:3000.
- To use this server in your frontend app: replace the backend url with: `"http://127.0.0.1:3000/"`

**NOTE:** Currently only get requests are working

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/devServer/package-lock.json
+++ b/hyrisecockpit/frontend/devServer/package-lock.json
@@ -1,0 +1,2071 @@
+{
+    "name": "debugServer",
+    "version": "1.0.0",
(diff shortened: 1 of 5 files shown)


MaxSchneider1337 (approved):
saves my day

PR #517, opened 2020-04-14
F/511 reformat line chart

### LINKED ISSUE #511: reformat line charts
The line chart layout needs an upgrade.

Here are some temporarily requirements, (which can be extended):
- shrink chart height
- add background color to comparison line chart (database color)
- move tile headers (database, title, metric details)
- remove x axis label (?)

**Affected Component(s):**  
`Frontend`

MaxSchneider1337:
# Resolves #511

improve design for the linecharts.

zero

**Affected Component(s):**  
Frontend, Linechart, Tile 

**Additional context:**  
zero

### DIFF
--- a/hyrisecockpit/frontend/src/components/charts/Linechart.vue
+++ b/hyrisecockpit/frontend/src/components/charts/Linechart.vue
@@ -125,7 +125,7 @@ function useLineChartConfiguration(
     return {
       xaxis: {
         title: {
-          text: props.chartConfiguration.xaxis
+          // text: props.chartConfiguration.xaxis
         },
         type: "date",
         tickformat: "%H:%M:%S",
@@ -143,7 +143,7 @@ function useLineChartConfiguration(
       margin: {
         l: 70,
         r: 40,
-        b: 70,
+        b: 30,
         t: 10,
         pad: 0
       },
--- a/hyrisecockpit/frontend/src/components/charts/Linechart.vue
+++ b/hyrisecockpit/frontend/src/components/charts/Linechart.vue
@@ -158,6 +158,7 @@ function useLineChartConfiguration(
     return {
       y: data,
       mode: "lines+markers",
+      fill: 'tonexty',
       line: database ? { color: database.color } : {},
       name: database ? database.id : {}
     };
--- a/hyrisecockpit/frontend/src/components/charts/Linechart.vue
+++ b/hyrisecockpit/frontend/src/components/charts/Linechart.vue
@@ -156,7 +156,7 @@ function useLineChartConfiguration(
     return {
       y: data,
       mode: "lines+markers",
-      fill: "tonexty",
+      fill: multipleDatabasesAllowed || "tonexty",
       line: database ? { color: database.color } : {},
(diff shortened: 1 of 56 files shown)


MaxSchneider1337:
this is how it looks like
image

hyrisecockpit/frontend/src/components/charts/Linechart.vue
```diff
@@ -158,6 +156,7 @@ function useLineChartConfiguration(
     return {
       y: data,
       mode: "lines+markers",
+      fill: "tonexty",
```
schTi: Color should only be filled on comparison page, right?

schTi (changes requested):

hyrisecockpit/frontend/src/components/charts/Linechart.vue
MaxSchneider1337: looks great on overview as well

schTi: Okay, can you add a screen shot of it?

Bouncner: And can you please check how that looks with more than 3 databases? Sample images from anywhere would be sufficient for me.

MaxSchneider1337: compare: 
image

overview: 
image

MaxSchneider1337: image
4 dbs break the format, but the plots seem to be too small anyway. I would suggest limiting the number of dbs on the comparison site to 3.

Bouncner: I agree that 4 is quite a lot, but the formatting should not be an issue with "a responsive" design, shouldn't it?

For the overview, I think the default line chart is a better fit.

MaxSchneider1337: 4 dbs also breaks the system info
image

MaxSchneider1337: @Bouncner sadly even most responsive designs break on a minimal with. We could also limit the max dbs according to the screen with. There should be enough space on a 4k screen, for 4 dbs.

Bouncner: I don't expect you to fit 17 databases into a single row. But the last picture of the instance view showed a case which should be managable when the database button, the title, and the metric each get their own row (or 2 and 1 on two rows). But that's not super important.

schTi (approved):
I am still not a fan of the idea to move the metric title to the left, but it is okay for me now. Nice design upgrade in general! :fire:

hyrisecockpit/frontend/src/components/charts/Linechart.vue
MaxSchneider1337: @Bouncner I wrote an extra issue for that. I tried to implement a simple responsive design, but it seems to be a bit more complex.

PR #519, opened 2020-04-15
f/510-add-footprint-chart

### LINKED ISSUE #510: add memory footprint line chart
Currently it is hard to notice changes in memory footprint over time. 

Thats why we want to add a new line chart containing information of the current memory footprint of all databases. This will add the ability to watch older footprint data, when adding the possibility to change the time range.

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #510

This PR adds a memory footprint chart, and ensures only fetching an endpoint once, also when multiple metrics have the same endpoint.

**NOTE:** There is no historical data for the storage endpoint yet

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/config.ts
+++ b/hyrisecockpit/frontend/config.ts
@@ -1,7 +1,10 @@
 const vm = "aurora";
 const port = "8000";
-//const backendUrl = `http://vm-${vm}.eaalab.hpi.uni-potsdam.de:${port}/`;
+// const backendUrl = `http://vm-${vm}.eaalab.hpi.uni-potsdam.de:${port}/`;
 const backendUrl = "http://127.0.0.1:3000/";
 
 export const monitorBackend = backendUrl + "monitor/";
 export const controlBackend = backendUrl + "control/";
+
+export const isInTestMode: boolean =
+  (window as any).Cypress || backendUrl.includes("127.0.0.1");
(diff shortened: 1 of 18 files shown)


hyrisecockpit/frontend/config.ts
```diff
@@ -1,6 +1,10 @@
 const vm = "aurora";
 const port = "8000";
 const backendUrl = `http://vm-${vm}.eaalab.hpi.uni-potsdam.de:${port}/`;
+//const backendUrl = "http://127.0.0.1:3000/";
```
MaxSchneider1337, line 4: do you want to push this comment?

hyrisecockpit/frontend/src/controller/metricController.ts
```diff
@@ -45,12 +46,34 @@ export function useMetricController(): MetricController {
 
   function setupServices(): Record<Metric, MetricService> {
     const services: any = {};
-    availableMetrics.forEach(metric => {
-      services[metric] = useMetricService(metric);
+    getMetricsByEndpoint(availableMetrics).forEach(metrics => {
```
MaxSchneider1337, line 49: so this function returns all metrics that are return by an endpoint?

MaxSchneider1337 (approved):
good changes

hyrisecockpit/frontend/config.ts
schTi: yes, because I dont want to add this all the time :D

hyrisecockpit/frontend/src/controller/metricController.ts
schTi: yes it returns an array of arrays where all metrics with the same endpoint are inside the same array.

PR #523, opened 2020-04-15
F/489 change distinctness of databases on comparison page

### LINKED ISSUE #489: improve database distinctness on comparison-screen
Databases are hard to differentiate in the comparison screen.
 
Implement a colored header (should be fixed on top of the screen) and draw a vertical line between db instances.

**Additional context:**  
image

Before: (please ignore the long database names for now)
image

After:
image

monasobh:
# Resolves #489 

It adds distinctness to the databases. 

**Affected Component(s):**  
`Frontend`

**What needs to be done?**
- [x] delete margin top from access freq. metric
- [x] make the database information box responsive when adding many databases
- [x] change font of y-axis to Arial (like others)
- [x] give database name more space than numbers
- [x] reduce space between host and name

**Whats new?**
Please be aware that the names und numbers are mocked, to use them as placeholders.

Comparison Page before: 
comparison_before

1. Deleted "Database Systems Information" Panel Header, taking too much space
2. Added color for Database information background
3. add vertical lines (border) in database color for better distinctness while scrolling 
4. Made it a one-liner. (I don't think it makes a lot of sence to be able to collapse the box. It wouldn't be much more space we would gain and there would need to be a button somewhere to indicate that all boxes would collapse at the same time - otherwise the metrics wouldn't be on the same height). 
~5. deleted round database names in metrics and left it only on the first metric~
6. Change font-style of cockpit to regular and color to dark grey (our secondary standard color)
7. Align Cockpit Navbar Header with colored database information boxes
8. delete space between database information box and first metric of database column 
9. add more space between databases
10. change: Memory Capacity --> Data Size, Memory Footprint --> Process Memory (RSS)
~11. change opacity to 90%~

0705_more margin

temp

temp3

column bottom border: 
image


monasobh:
@Bensk1 @Bouncner what do you think about the design? Could you please give me some feedback.

Bensk1:
I like it, especially 1+4 and 3. Since this was Martin's idea, he should have the final `go` here.

Bouncner:
Looking good. A few small remarks:
 * I wouldn't put the description (host, capacity) below the actual value
 * I wouldn't put the units in bold, only the database name
 * Why does `database-1` and `Markets` differ?
 * What happens when we scroll? I am not a big fan of floats, so I think the "database bubble" would be helpful in any chart as long as we don't need the space (which is the case right now).
 * Can we somehow connect the "titlebox" with the vertical lines besides the plots?
 * I think half the width of the vertical lines is still fine.

When we settled on these remarks, we should have a little discussion round with all team members.

monasobh:
Please checkout the changes I made and let me know, what you think.

Bouncner:
If the font needs to be bold in order to be readable, we have either too dark background or we need to switch to white for certain colors.

Here is an algorithm that checks for the perceived contrast. This way, we could switch to white whenever necessary. But I am not sure if we want a database with white text on the left and another one with black text next to it.
Any ideas here?

Besides that, I like the PR. Maybe a bit more white space around the databases.

monasobh:
>Here is an algorithm that checks for the perceived contrast. This way, we could switch to white whenever necessary. But I am not sure if we want a database with white text on the left and another one with black text next to it.
> Any ideas here?

I think white is the only option here. Couldn't find another color that is readable on all database colors. What do you think (new photo above)?

Bouncner:
Argh, I did not paste the link: https://stackoverflow.com/a/3943023/1147726 (and the linked duplicate issue).

The white wouldn't work for the light blue in the first pictures, wouldn't it? It this light blue gone?
Can you show the white text for all the possible colours?

monasobh:
> Argh, I did not paste the link: https://stackoverflow.com/a/3943023/1147726 (and the linked duplicate issue).
> 
> The white wouldn't work for the light blue in the first pictures, wouldn't it? It this light blue gone?
> Can you show the white text for all the possible colours?

Yes, the light blue is gone and replace by a stronger one. See above for all databases.

hyrisecockpit/frontend/src/components/container/MetricTile.vue
```diff
@@ -1,6 +1,14 @@
 <template>
-  <v-card :id="`${tileDatabase}-${metric}-tile`">
-    <v-card-title class="card-title" :ref="`${tileDatabase}-${metric}-title`">
+  <v-card
+    class="card"
+    :style="getBorderColor(tileDatabase)"
+    :id="`${tileDatabase}-${metric}-tile`"
```
schTi: please move the id to the top

hyrisecockpit/frontend/src/components/container/MetricTile.vue
```diff
@@ -49,6 +57,7 @@ import {
 import { Database } from "../../types/database";
 import DatabaseChip from "@/components/details/DatabaseChip.vue";
 import TimeInterval from "@/components/details/TimeInterval.vue";
+import { Color } from "plotly.js";
```
schTi: this import isnt used anywhere

hyrisecockpit/frontend/src/components/container/MetricTile.vue
```diff
@@ -113,4 +140,7 @@ export default defineComponent({
   align-items: flex-start;
   justify-content: flex-end;
 }
+#comparison-page .card {
```
schTi, line 142: Why do you need this class only on comparison page? Couldnt we have an equal design on the overview page?

hyrisecockpit/frontend/src/components/container/MetricsComparisonTable.vue
```diff
@@ -34,15 +40,24 @@ import {
 import { ContainerProps, ContainerPropsValidation } from "../../types/views";
 import { useDatabaseFlex } from "../../meta/components";
 import MetricTile from "@/components/container/MetricTile.vue";
+import DatabaseSystemDetails from "../../components/details/DatabaseSystemDetails.vue";
```
schTi: unnecessary import here

hyrisecockpit/frontend/src/components/details/DatabaseSystemDetails.vue
```diff
@@ -110,25 +96,37 @@ export default defineComponent({
...
+#comparison-page .col,
+#overview-page .col {
+  padding: 0px;
+}
+#comparison-page .card,
+#overview-page .card {
+  display: flex;
+  flex-direction: column;
+  margin-right: 6px;
+  margin-left: 6px;
```
schTi: the db system details are only shown on comparison and on overview page, so imo it is not necessary to scope those css classes here

schTi (changes requested):

hyrisecockpit/frontend/src/components/container/MetricTile.vue
monasobh: Unfortunately not. The V-card on the comparison page can't be round on all corners, otherwise the vertical lines from the databases won't align right.

hyrisecockpit/frontend/src/components/details/DatabaseSystemDetails.vue
monasobh: If I don't use the #, all the v-cards we use e.g. Plugins will change..

hyrisecockpit/frontend/src/components/container/MetricTile.vue
schTi: you can make the css scoped, just use `<style scoped>` instead of `<style>`

hyrisecockpit/frontend/src/components/details/DatabaseSystemDetails.vue
schTi: you can make the css scoped, just use <style scoped> instead of <style>

hyrisecockpit/frontend/src/components/container/MetricTile.vue
schTi: nvm wrong comment section

schTi (approved):
Thanks for adding requested changes. Feel free to merge!

PR #524, opened 2020-04-15
U/docker development environment

### LINKED ISSUE #526: Alternative vm independent developer set-up
In the last few days, we had some issues working with the VM's. Moreover, we often collide with the usage of the hyrise databases.

I think a good solution would be to work with docker as a developer tool. The Image would be based on Ubuntu and comes with a hyrise, influxdb, and all other packages. So we can simply pull the image, mount our working directory, and get via a sup-shell inside. From there we could commit and push/pull. 

I think to be flexible we could use the following three containers: 

Screenshot 2020-04-16 at 12 08 30

Screenshot 2020-04-16 at 12 16 44

Important to mention for mac and windows user is that docker needs a Linux kernel to run. So, as a result, it needs to be virtualized. The virtualization that is used is like a host virtualization. This kind is not very performant. Because if we leave/enter the container the host kernel is getting restructured.
It's kind like a teenager's house party. If the parents are coming to the house gets cleaned up. If they are leaving the house gets messy again. 
So for development on Mac/windows, it is better to leave the container as rare as possible. To have the components in one container is an advantage. 
Another advantage is, if we are working on the VM everyone is isolated. Moreover, all the important packages are already ready and we can start right away. 

I think it is important to mention that everyone should use the tools where they are productive. If we are working on a VM (especially as backend developers) we could just simply use a sheet to give every person some ports they can use. 
If the VM 's are not working the docker set-up would be a nice alternative to quickly continue development.   

I considered running everything local. I think as a ubuntu user that's a good way to go.  It's not difficult to set up a hyrise. But as a Mac it is not possible at the moment to run the cockpit directly. To run multiple containers on Mac isn't a good Idea as well (like mentioned above).

Alexander-Dubrawski:
# Resolves #526 
In the last few days, we had some issues working with the VM's.

 I created three docker images based on ubuntu that are hopefully an alternative to working directly on the VM'S. The three images are:

* `cockpit_minimal` it just gives us a simple environment with all packages for python and javascript

The idea is to mount our working directory and work inside the ubuntu system of the container. So we will git pull and commit from inside the container.

Screenshot 2020-04-16 at 12 08 30

More Information on how to use this set-up will be in the read me.
A nice side effect of the containers is that we can develop on every platform and isolated from each other. What would be helpfull on the VM's.

**Additional context:**  
I think it is still very important to use the tools you prefer. Everyone has there set-up where they can develop the best. The goal of this PR is to give you some alternatives that can help in some situations. 

## Todo 

- [x] write a readme that explains how to use the image 
- [x] write a python script that starts all the backend component (cockpit, hyrise, influx)
- [x] build final image versions


Bouncner:
What are the costs of not running Hyrise and InfluxDB in the minimal setup?

And what is the difference between Udo's image and the complete image?

Alexander-Dubrawski:
> What are the costs of not running Hyrise and InfluxDB in the minimal setup?

We need to differentiate between Mac OSX and Linux. On Linux, you can run Hyrise and InfluxDB outside of the container and access it from inside. Of course, you need to take care that you have the right hyrise and influx version and you need to update them. The image is taking care of that for you.
For mac and windows docker needs a Linux kernel to run. So, as a result, it needs to be virtualized. The virtualization that is used is like a host virtualization. This kind is not very performant. Because if we leave/enter the container the host kernel is getting restructured.
It's kind like a teenager's house party. If the parents are coming to the house gets cleaned up. If they are leaving the house gets messy again.
That's why a multi-container set up for mac is pretty bad. 
On my mac, I had a lot of problems running the components in different containers. 

> And what is the difference between Udo's image and the complete image?

The main difference is intention. I created these containers to use them as a development environment. On the other side, Udo uses images for every single component and the main purpose is to execute the `.yml` file which starts every single component in a separate container. So here is the goal to start the environment.  The images I wrote have the goal to give you a ubuntu environment where you can develop inside (git,... is all installed). For example, you can open as many subshells you want to these containers and inside you can commit and push. In conclusion, the main purpose is to create a ubuntu development environment with all packages installed.   

You can start all components in the complete image as well. But if that is your purpose and you are working on local ubuntu I would recommend using Udo's script. If you are working on the mac on the other hand or you are forced to work on a VM (with other developers) you could use this image (in the case of the VM an excel sheet to give all the developer ports would be also fine).

Bouncner:
I don't get this metaphor with the parents house. Didn't like these parties anyways, always cheap beer.

@cH3n7i ... this aspect with the messy containers that don't work well on Alex's laptop. Can we address this with a single docker image?
I don't like this development of having several docker setups.

Alexander-Dubrawski:
> I don't get this metaphor with the parents house. Didn't like these parties anyways, always cheap beer.
> 
> @cH3n7i ... this aspect with the messy containers that don't work well on Alex's laptop. Can we address this with a single docker image?
> I don't like this development of having several docker setups.

I think the docker set up from Udo for deployment is fine. The Idea of docker with deploying software is to separate the logical component. I think we could put all three backend components in one image but the rest is totally fine. On Linux and more powerful macs, it should work with no big problems. I think it is okay to have a docker development set up and a deployment one. I wouldn't deploy all components in one image (that just doesn't feel right and the purpose is not to develop). Since we will have mac support and if the mac you will run it on can't handle it, there will be an option to install it directly.

cH3n7i:
> For mac and windows docker needs a Linux kernel to run. So, as a result, it needs to be virtualized. The virtualization that is used is like a host virtualization. This kind is not very performant. Because if we leave/enter the container the host kernel is getting restructured.
> It's kind like a teenager's house party. If the parents are coming to the house gets cleaned up. If they are leaving the house gets messy again.
> That's why a multi-container set up for mac is pretty bad.

With the new `Docker.app` we're actually using the Hypervisor Framework (introduced in 10.10, ~ 5 years ago), which should eliminate a lot of the overhead usually found in VM-Environments.
see: https://docs.docker.com/docker-for-mac/faqs/ 
(FYI: We currently have most of our Macs using the containerized solution running fine.)
 
> On my mac, I had a lot of problems running the components in different containers.

If you have some of the points on this, I'd love to fix them for the "separated setup".  

> You can start all components in the complete image as well. But if that is your purpose and you are working on local ubuntu I would recommend using Udo's script. If you are working on the mac on the other hand or you are forced to work on a VM (with other developers) you could use this image (in the case of the VM an excel sheet to give all the developer ports would be also fine).

While both setups work fine on Unix systems (e.g. our Ubuntu server or Ubuntu and macOS clients) - the containerized (one component in one container) aims to have more modularity and mimic the server. Development happens on the host, the container will just execute the magic.

Bouncner:
Don't get me wrong when I am sounding very critical. It might be my very limited understanding of Docker. But I want to make sure that we do not unnecessarily add anything that nobody will ever use again.
Hence, the place where I would see any Docker file that is not the main Docker file, is an additional repository.

> The Idea of docker with deploying software is to separate the logical component.

What do you mean?

> I think we could put all three backend components in one image but the rest is totally fine.

That does not really help if I still need to  manage version manually and connect components manually.

I still don't see the reason for the whole script. I still haven't heard numbers why we should not run InfluxDB and Hyrise within the Docker image.
For my taste, that whole discussion is too unclear. There should be strong reasons for additional images and I haven't heard a good reason yet.

Bouncner:
I guess I need to better understand what both's aproaches disadvantages are. When @Bensk1 is ready with catching up, maybe have a devops call for half an hour to explain that all to me? :)

Bensk1:
How about talking about that during (or probably better after) today's meeting?

utils/docker_developent_enviroment/complete/install_dependencies.sh
```diff
@@ -0,0 +1,91 @@
...
+if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
+    unamestr=$(uname)
+    if [[ "$unamestr" == 'Darwin' ]]; then
+        brew --version 2>/dev/null || /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+
+        echo "Installing dependencies (this may take a while)..."
+        if brew update >/dev/null; then
+            # python2.7 is preinstalled on macOS
+            # check, for each program (aka. formula) individually with brew, whether it is already installed due to brew issues on MacOS after system upgrade
+            # NOTE: The Mac CI server does not execute the install_dependencies.sh - formulas need to be installed manually.
```
Bensk1: Is this a complete copy of the Hyrise install dependency script?

Alexander-Dubrawski: Nearly. I needed to remove the first part because docker wasn't able to locate `/dev/tty` in:

https://github.com/hyrise/hyrise/blob/9482e543b4082d19c594e21fc7765d8108caf506/install_dependencies.sh#L12

Moreover, I hade to remove `sudo` in the script.

Bensk1: Alright. We adapt the install script from time to time so this will be another thing you have to maintain. I don't know if it's wise to have an almost exact copy of it.

Bouncner:
I know it: it’s not. :-)

Alex, you should definitely get the script (curl/wget) and execute the most current script.
Do you have an idea how to avoid this dev problem? We could adapt the Hyrise script, so that you could use the script without modifications.

utils/docker_developent_enviroment/complete/install_dependencies.sh
Alexander-Dubrawski: As discussed in the Meeting I removed the two dockerfiles that are building the two bigger images. So there will be no Hyrise in the image anymore and as a result of that, the script is not needed anymore.

Alexander-Dubrawski:
> I know it: it’s not. :-)
> 
> Alex, you should definitely get the script (curl/wget) and execute the most current script.
> Do you have an idea how to avoid this dev problem? We could adapt the Hyrise script, so that you could use the script without modifications.

As discussed in the Meeting I removed the two dockerfiles that are building the two bigger images. So there will be no Hyrise in the image anymore.

Alexander-Dubrawski:
I will put the Dockerfiles in a separate repro.

PR #527, opened 2020-04-16
Docker/prepare multiple instance one host

cH3n7i:
Currently just one instance of docker container can be run. Since we're far away from performance limits, we should be able to run multiple instances for improved workflows and stability.

We slim down the images, the frontend distribution and prepare the Dockerfiles for proxy use.
Additionally InfluxDB will be auto-resetted on restart, removing unwanted manual steps.

**Affected Component(s):**  
`Docker`, `InfluxDB`


heseding (approved):
We should add a pattern for `Dockerfile` to `CODEOWNERS` as well.

PR #531, opened 2020-04-16
F/512-improve-ux-on-workload-generation

### LINKED ISSUE #512: improve ux on workload generation
It is not always visible why you can not load or start a workload.

Adding more user instructions could help.

**Affected Component(s):**  
`Frontend`

### LINKED ISSUE #538: show message when hyrise is not active
Currently we dont see, when a hyrise instance is not active. We definitely need an indicator for this. Design suggestions are welcome!

**Affected Component(s):**  
`Frontend`

kathwill:
# Resolves #512 #538 

It is not always visible why you can not load or start a workload. 

There are now more warnings to inform the user and all user input on the workload generation dialog is disabled when no database is added, an instance is blocked or a workload is loaded.

**Affected Component(s):**  
`Frontend`

NoDatabaseAdded

WarningChipWithIcon


Bouncner:
What needs to happen that a database is "blocked"?

kathwill:
> What needs to happen that a database is "blocked"?

It is possible that a database is blocked when loading or removing data and when starting or stopping a workload. So now it is visible why certain buttons are disabled and which of the instances is still blocked. 
Do you think "blocked" is the wrong word for that?

Bouncner:
No, I think “blocked” is ok. Just wasn’t sure what it means.

Would it be possible to know why it is blocked? (e.g. “blocked due to data loading”)

hyrisecockpit/frontend/src/components/alerts/StatusWarning.vue
```diff
@@ -0,0 +1,71 @@
...
+    </warning>
+    <warning v-for="database in inactiveDatabases" :key="database">
+      <template #message> {{ database }} is not active </template>
+    </warning>
+    <v-row>
+      <v-chip
+        class="mx-4 mt-4 mb-1 py-2 white--text"
+        v-for="database in blockedDatabases"
+        :key="database"
+        :color="useDatabaseService().getDatabaseColor(database)"
```
schTi: It is a bit overkill to import the big database service here to just get the color. You can access those colors via the database controller directly, because they are stored together with the databases itself. You can have a look at the `DatabaseSystemDetails.vue`.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -41,17 +42,17 @@
...
               >
               </v-radio>
             </v-radio-group>
             <v-btn-toggle>
               <v-btn
                 id="start-workload"
                 @click="handleButtonChange('start')"
-                :disabled="buttons.start.loading || !isLoaded(workload)"
+                :disabled="
+                  buttons.start.loading || !isLoaded(workload) || isDisabled()
```
schTi: Currently the tests are failing, because the buttons are not getting disabled, when there is no database. Thats because the `isDisabled` function here will only be called when mounting the component. To get notified, when there are changes, you have to use a computed property instead.

hyrisecockpit/frontend/src/meta/events.ts
```diff
@@ -88,11 +94,23 @@ export function useDatabaseEvents(): {
...
   }
+  function emitNoDatabaseEvent(): void {
+    eventBus.$emit("NO_DATABASE");
+  }
+  function emitDatabaseStatusEvent(
+    database: string,
+    blocked: boolean,
+    active: boolean
+  ): void {
+    eventBus.$emit("DATABASE_STATUS", database, blocked, active);
```
schTi: can you please rename this to `DATABASE_STATUS_CHANGED`? This would be more descriptive.

hyrisecockpit/frontend/src/meta/events.ts
```diff
@@ -88,11 +94,23 @@ export function useDatabaseEvents(): {
   function emitDatabaseRemovedEvent(): void {
     eventBus.$emit("DATABASE_REMOVED");
   }
+  function emitNoDatabaseEvent(): void {
+    eventBus.$emit("NO_DATABASE");
```
schTi: Same here. Maybe `NO_DATABASE_ADDED` would be better, to distinguish between added and selected.

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -34,6 +35,41 @@ export function useDatabaseController(): DatabaseController {
...
+      if (!active && !inactiveDatabases.includes(database)) {
+        inactiveDatabases.push(database);
+      } else if (active && inactiveDatabases.includes(database)) {
+        let index: number = inactiveDatabases.indexOf(database);
+        inactiveDatabases.splice(index, 1);
+      }
+    }
+  );
+
+  function getDatabaseStatusInformation(): DatabaseStatus {
```
schTi: When I said, we can store those status information directly with the other database information, I really meant this. Why dont you just add `blocked` and `active` as additional properties to the database data inside the getDatabaseInformation- function. Imo you can initialize those values with `false` or `true`. Because the `status` endpoint will be called every second anyway.

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -34,6 +35,41 @@ export function useDatabaseController(): DatabaseController {
...
+  let noDatabase: boolean = true;
+  let blockedDatabases: string[] = [];
+  let inactiveDatabases: string[] = [];
+
+  eventBus.$on("NO_DATABASE", () => {
+    noDatabase = true;
+  });
+
+  eventBus.$on(
+    "DATABASE_STATUS",
```
schTi: When you apply the changes of my other comment, then you can get rid of many of these lines. It should be something like: `const database = getDatabaseById(databaseId)` and then just adapt the entries `database.active` and `database.blocked` accordingly to the new values.

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -34,6 +35,41 @@ export function useDatabaseController(): DatabaseController {
     }
   });
 
+  let noDatabase: boolean = true;
+  let blockedDatabases: string[] = [];
+  let inactiveDatabases: string[] = [];
+
+  eventBus.$on("NO_DATABASE", () => {
```
schTi: The database controller stores all running databases. So he already knows, if there isnt any database.

schTi (changes requested):

hyrisecockpit/frontend/src/components/alerts/StatusWarning.vue
```diff
@@ -0,0 +1,71 @@
...
+  useDatabaseService: () => void;
+}
+
+export default defineComponent({
+  name: "StatusWarning",
+  components: {
+    Warning,
+  },
+  setup(props: {}, context: SetupContext): Data {
+    const noDatabase = ref<boolean>(false);
```
MaxSchneider1337: could you please rename `noDatabase` to something more descriptive like `noDatabaseAvailable`. It is kind of hard to now if this is a boolean.

MaxSchneider1337: just read @schTi s comments and he suggested `selected` and `selected`. I don't mind the actual naming, but it should be more descriptive.

hyrisecockpit/frontend/src/components/alerts/StatusWarning.vue
```diff
@@ -1,23 +1,24 @@
...
-        <v-icon> mdi-exclamation </v-icon>
-        <span> {{ database }} is blocked </span>
-      </v-chip>
+      <div v-for="database in databases" :key="database.id">
+        <warning class="mx-4" v-if="!database.active">
+          <template #message> {{ database.id }} is not active </template>
+        </warning>
+        <v-chip
+          class="mx-4 mt-4 mb-1 py-2 white--text"
+          v-if="database.blocked"
```
schTi, line 24: just move the v-if directive to the top please to improve readability.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -42,7 +42,9 @@
                 :key="workload"
                 :label="getDisplayedWorkload(workload)"
                 :value="workload"
-                :disabled="!isLoaded(workload) || isDisabled()"
+                :disabled="
+                  !isLoaded(workload) || isDisabled() || !databases.length
```
schTi: Can you tell me why you did not make the `isDisabled` a computed property?

schTi (changes requested):

hyrisecockpit/frontend/src/components/workload/WorkloadActions.vue
```diff
@@ -0,0 +1,82 @@
...
+      type: Object,
+      default: null,
+    },
+    disabled: {
+      type: Boolean,
+      default: null,
+    },
+  },
+  setup(props: Props, context: SetupContext): Data {
+    function startingWorkload(): void {
```
schTi: You dont need those functions to just emit an event. You can directly do this in the template. For example: `@click="$emit('start')"`. You can do this everywhere when there is no other logic.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -9,110 +9,41 @@
         <v-icon @click="closeWorkloadDialog()">mdi-close</v-icon>
       </v-system-bar>
       <v-card-text>
+        <status-warning
+          :selected-databases="databases"
+          :selected-metrics="['']"
```
schTi, line 14: why not pass an empty array here?

hyrisecockpit/frontend/src/components/workload/WorkloadSelector.vue
```diff
@@ -0,0 +1,52 @@
...
+  getDisplayedWorkload: (workload: Workload) => string;
+  handleWorkloadChange: (workload: Workload) => void;
+}
+
+export default defineComponent({
+  name: "WorkloadSelector",
+  props: {
+    workloadData: {
+      type: Object,
+      default: null,
```
schTi: you should always set a good default value. When you set null here, there will be an error, when no workload data has been passed. An empty object would be  the better choice for sure. For boolean values you should always take `true` or `false` as default. So you just need to pass a value in the other case.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -9,110 +9,41 @@
...
-                <v-icon>
-                  mdi-stop
-                </v-icon>
-              </v-btn>
-            </v-btn-toggle>
+            <frequency-handler @change="handleFrequencyChange($event)" />
+            <workload-selector
+              :workload-data="workloadData"
+              :disabled="disabled"
+              @change="handleWorkloadChange($event)"
```
schTi:
```suggestion
              @change="handleWorkloadChange"
```
Thats exactly the same, but prettier :smile:

schTi (changes requested):
I really like the changes you made. The template is more readable now. But Im still having issues with the script part of the workload generation component. The setup functions is way to big. Its hard to see whats going on there, and what belongs to each other. Thats exactly what we want to prevent by using vue composition api. To increase the code organisation, I would suggest to split all related logic into an own function. For example: `useWorkloadActions` should contain everything needed for handling workload related stuff (starting, stopping, ..). These functions should be placed outside the setup function and they should have an own data interface. The setup function should then just return these functions. If you have any questions about that, just hit me up.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: Then the warning will always be shown because there is no metric

hyrisecockpit/frontend/src/components/workload/WorkloadActions.vue
kathwill: Thanks!

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: Thanks!

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
schTi: Ah, I see.

kathwill: Do you have a better way of doing that or should I just leave it as it is?

schTi: Its okay for me

hyrisecockpit/frontend/src/types/workloads.ts
```diff
@@ -11,3 +13,18 @@ export const availableWorkloads: Workload[] = [
   "tpcds",
   "job",
 ];
+
+export interface WorkloadAction {
```
schTi: you can move this interface to the workload generation as well, because we dont need this anywhere else yet

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -103,201 +92,199 @@ export default defineComponent({
...
-      },
-      tpcds: {
-        loaded: false,
-        loading: false,
-      },
-      job: {
-        loaded: false,
-        loading: false,
-      },
+  setup(props: {}, context: SetupContext): WorkloadAction {
```
schTi: thats not what I expected. You just moved all the stuff to an additional function. What I meant was splitting it in different functions. E.g. one function handles all frequency related stuff, the other one handles all loading stuff, ... . With your solution we still have the same readability issues.
Or do you think it is not possible to split this like I suggested?

schTi (changes requested):

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: It's not possible to split workload and frequency, but workload-data is possible. So I will do that now

Bouncner:
Can you change the title to "Workload Settings" which is consistent with the menu?

kathwill:
> Can you change the title to "Workload Settings" which is consistent with the menu?

Yes

hyrisecockpit/frontend/src/controller/databaseController.ts
```diff
@@ -37,9 +37,11 @@ export function useDatabaseController(): DatabaseController {
   eventBus.$on(
     "DATABASE_STATUS_CHANGED",
     (databaseId: string, blocked: boolean, active: boolean) => {
-      const database = getDatabaseById(databaseId);
-      database.active = active;
-      database.blocked = blocked;
+      if (databasesUpdated.value) {
```
schTi, line 40: Why do you need this?

schTi (commented):
Thanks for the great work, just one question.

hyrisecockpit/frontend/src/controller/databaseController.ts
schTi: Okay nvm, I got it :smile:

schTi (approved):

hyrisecockpit/frontend/src/controller/databaseController.ts
kathwill: In the setup there was sometimes an error because getDatabaseById(databaseId) was not defined yet

PR #532, opened 2020-04-16
B/Add mac OS X support

Alexander-Dubrawski:
With the threaded connection pool from psycopg2, we weren't able to run the backend native on mac.

We can now run the backend on the mac 🥳 .

**Affected Component(s):**  
`Background_Scheduler`, `Worker_pool`, `Queue`

**Additional context:**  
The support for multiprocessing of python on a mac is a little bit strange. I needed to write a small wrapper for the `Queue` that is using a lock to access it. With this lock, the queue acts the same as on Linux. Moreover, I needed to write a wrapper for `Process` in the parallel table loading logic. In Mac osx it is not possible to give `Process` a  function target that lies in the same class. My solution at the moment is a little bit hacky. Last but not least I needed to write a wrapper around the MagicMock as well, for the context of using it for multiprocessing testing.


hyrisecockpit/database_manager/worker_pool.py
```diff
@@ -1,13 +1,19 @@
...
 from apscheduler.schedulers.background import BackgroundScheduler
 
 from .cursor import ConnectionFactory
 from .worker import execute_queries, fill_queue
 
+if platform.startswith("linux"):
+    from multiprocessing import Event, Process, Queue, Value
+elif platform.startswith("darwin"):
+    from multiprocessing import Event, Process, Value
+    from hyrisecockpit.cross_platform_support.mac_osx_support import MacQueue as Queue  # type: ignore
```
heseding: How about we import platform-specific components like this:
```suggestion
from multiprocessing import Event, Process, Value
from hyrisecockpit.cross_platform_support.mac_osx_support import Queue
```
and handle the if/else in the `mac_osx_support.py` file?
This way, we would only declare platform specifics once in a single file.

hyrisecockpit/cross_platform_support/mac_osx_support.py
```diff
@@ -0,0 +1,115 @@
...
+    formatted_parameters = format_query_parameters(parameters)
+    try:
+        with connection_factory.create_cursor() as cur:
+            cur.execute(query, formatted_parameters)
+            success_flag.value = True
+    except (DatabaseError, InterfaceError, ProgrammingError):
+        return None  # TODO: log error
+
+
+class LoadingTabbleProcess:
```
heseding:
```suggestion
class LoadingTableProcess:
```

hyrisecockpit/database_manager/worker_pool.py
Alexander-Dubrawski: I like the idea. How about to rename `mac_osx_support ` to `cross_platform_support` and import all cross-platform critical modules from there?

PeterTsayun (approved):
Well done! :thumbsup:

PR #542, opened 2020-04-21
Refactor sql namespace

Alexander-Dubrawski:
This Pr continues the Flask backend refactoring for the SQL namespace api. 

**Affected Component(s):**  
`App`

## Todo 

- [x] Write tests for model
- [x] Write tests for interface
- [x] Write tests for service
- [x] Write tests for controller
- [x] Write tests for schema


hyrisecockpit/api/app/sql/model.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, id: str, query: str):
+        """Initialize a sql query model."""
+        self.id: str = id
+        self.query: str = query
+
+    def update(self, interface: SqlQueryInterface) -> "SqlQuery":
+        """Update attributes of a sql query model."""
+        for key, value in interface.items():
+            setattr(self, key, value)
+        return self
```
heseding: Where is this method used? I suggest removing it.

hyrisecockpit/api/app/sql/model.py
```diff
@@ -0,0 +1,44 @@
...
+        self.successful: bool = successful
+        self.results: List[List[str]] = results
+        self.col_names: List[str] = col_names
+        self.error_message: str = error_message
+
+    def update(self, interface: SqlResponseInterface) -> "SqlResponse":
+        """Update attributes of a sql reponse model."""
+        for key, value in interface.items():
+            setattr(self, key, value)
+        return self
```
heseding: Where is this method used? I suggest removing it.

hyrisecockpit/api/app/sql/model.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class SqlResponse:
+    """Model of a sql query response."""
+
+    def __init__(
+        self,
+        id: str,
```
heseding, line 19: It would make sense to me to send the executed query along with the response (otherwise, the frontend would have to cache that information):

```suggestion
class SqlResponse(SqlQuery):
    """Model of a sql query response."""

    def __init__(
        self,
        id: str,
        query: str,
```

hyrisecockpit/api/app/sql/service.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        if response["header"]["status"] == 200:
+            return (
+                SqlResponse(**response["body"]["results"]),
+                200,
+            )
+        else:
+            return (
+                None,
+                response["header"]["status"],
```
heseding: It's rather inexplicit which status codes can occur here, I suggest handling each explicitly.

hyrisecockpit/api/app/sql/controller.py
```diff
@@ -0,0 +1,32 @@
...
+from .service import SqlService
+
+api = Namespace("SQL", description="SQL execution.")
+
+
+@api.route("/")
+class Sql(Resource):
+    """Execute SQL query on database."""
+
+    @api.response(404, "Database not found.")
```
heseding, line 21: This is not the only status code possible here since the status codes from the `DatabaseManager` get passed along in the service.

hyrisecockpit/api/app/sql/service.py
Alexander-Dubrawski: We can only have the status codes 200 and 404. Do you suggest to use `elif ` to improve the readability of the code?

hyrisecockpit/api/app/sql/controller.py
Alexander-Dubrawski: Which other codes besides the flask responses are possible? I just found 404 and 200.

hyrisecockpit/api/app/sql/model.py
Alexander-Dubrawski: At the moment I don't see the benefits to return the query as well. It would just overload the response. The frontend needs to wait for the reply either way (So they can easily hold on to the query). I think it should be their responsibility to keep track of the query.

hyrisecockpit/api/app/sql/service.py
```diff
@@ -36,8 +37,12 @@ def execute_sql(
...
+        elif response["header"]["status"] == 404:
             return (
                 None,
-                response["header"]["status"],
+                404,
+            )
+        else:
+            raise StatusCodeNotFoundException(
+                f"Unknown status code: {response['header']['status']}"
             )
```
fabianhe:
```suggestion
            raise NotImplementedError()
```

heseding (approved):

PR #546, opened 2020-04-22
B/quick fix update plugin log

PeterTsayun:
# Resolves #number

Hyrise uses ms-precision timestamps for plugin logs. So, the query needs to be adjusted.
Moreover, the syntax for attribute access in Hyrise is `"attribute"` instead of `'attribute'`. 

No feature added.

**Affected Component(s):**  
`BackgroundScheduler`

**Additional context:**

### DIFF
--- a/hyrisecockpit/database_manager/background_scheduler.py
+++ b/hyrisecockpit/database_manager/background_scheduler.py
@@ -241,7 +241,7 @@ class BackgroundJobManager(object):
 
     def _update_plugin_log(self) -> None:
         """Update plugin log."""
-        log_df = self._sql_to_data_frame("SELECT * FROM meta_log;", None,)
+        log_df = self._sql_to_data_frame("SELECT * FROM meta_log;", None)
 
         if log_df.empty:
             return
--- a/hyrisecockpit/database_manager/background_scheduler.py
+++ b/hyrisecockpit/database_manager/background_scheduler.py
@@ -241,7 +241,13 @@ class BackgroundJobManager(object):
 
     def _update_plugin_log(self) -> None:
         """Update plugin log."""
-        log_df = self._sql_to_data_frame("SELECT * FROM meta_log;", None)
+        endts = int(time_ns() / 1_000_000)  # timestamps in hyrise are in ms-precision
+        startts = endts - 5_000
+
+        log_df = self._sql_to_data_frame(
+            """SELECT * FROM meta_log WHERE "timestamp" >= %s AND "timestamp" < %s;""",
+            params=(startts, endts),
+        )
 
         if log_df.empty:
             return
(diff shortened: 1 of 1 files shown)


Bouncner (commented):
What is the exact error?
We support time stamps “somehow” by using strings.

PeterTsayun:
> What is the exact error?
> We support time stamps “somehow” by using strings.

As mentioned in Slack, `SELECT * FROM meta_log WHERE 'timestamp' >= 100000;` leads the Hyrise instance to the crash:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  src/lib/expression/evaluation/expression_evaluator.cpp:1074 Argument types not compatible
Aborted
```

The timestamp is stored as an int in Hyrise, so we compared it with another UNIX-timestamp (100000 in the example above). What would be the supposed way to filter plugin logs in the `meta_log`-table based on the timestamps?

Bouncner:
`SELECT * FROM meta_log WHERE timestamp >= 100000;`

PeterTsayun:
> SELECT * FROM meta_log WHERE timestamp >= 100000;

This approach causes the following syntax error:
```
=> SELECT * FROM meta_log WHERE timestamp >= 100000;
Invalid input error: SQL query not valid.
SQL query:
==========
SELECT * FROM meta_log WHERE timestamp >= 100000;
                             ^=== ERROR HERE!
==========
Error line: 0
Error column: 29
Error message: syntax error, unexpected TIMESTAMP
```

Bouncner:
What's the result of  `SELECT * FROM meta_log`?

PeterTsayun:
> What's the result of `SELECT * FROM meta_log`?

It works just fine:

```
   timestamp   |     reporter     |                               message                                
---------------+------------------+----------------------------------------------------------------------
 1587566538185 | ClusteringPlugin | Initialized!
 1587566544493 | ClusteringPlugin | Applied new clustering configuration (l_shipdate) to lineitem table.
 1587566545824 | ClusteringPlugin | Applied new clustering configuration (o_orderdate) to orders table.
(3 rows)
```

Is `TIMESTAMP` a reserved keyword in Hyrise?

Bouncner:
I’ll take a look at it.

No, it shouldn’t be reserved.

Bouncner:
Oh wait, the parser might not like it. Can uou try `”timestamp”`?

PeterTsayun:
> Oh wait, the parser might not like it. Can uou try `”timestamp”`?

Yes, now it seems to be working:

```
=> SELECT * FROM meta_log WHERE "timestamp" > 1587566538185;

   timestamp   |     reporter     |                               message                                
---------------+------------------+----------------------------------------------------------------------
 1587566544493 | ClusteringPlugin | Applied new clustering configuration (l_shipdate) to lineitem table.
 1587566545824 | ClusteringPlugin | Applied new clustering configuration (o_orderdate) to orders table.
(2 rows)

```

Thanks for the hint! :blush:

Bouncner:
https://stackoverflow.com/a/13089425

Alexander-Dubrawski (approved):

PR #551, opened 2020-04-23
F/fix-workload-endpoints

kathwill:
# Resolves #number

Fixes workload requests and adds new endpoints

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
+++ b/hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
@@ -21,7 +21,7 @@
               thumb-label="always"
               min="0"
               max="1000"
-              @click="handleWorkloadChange()"
+              @click="handleFrequencyChange()"
             ></v-slider>
             <v-text-field
               id="frequency-field"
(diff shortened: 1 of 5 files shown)


hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -210,26 +214,47 @@ export default defineComponent({
...
-        });
-      } else {
-        stopWorkload().then(() => {
-          buttons[button].loading = false;
-          buttons[button].active = true;
-        });
-      }
+      getWorkloads().then((response: any) => {
+        if (response.data.length === 0) {
+          if (button === "start") {
```
MaxSchneider1337: this function definitely needs to be refactored. 
Since You already have 3 Buttons you could just write 3 functions, to avoid all the ifs. You can also write an extra function like `stopLoading` to reuse `buttons[button].loading = false; buttons[button].active = true;`

MaxSchneider1337 (commented):
thanks for fixing this

schTi (approved):
Thanks for the fix! In my opinion we still need to refactor the whole workload generation component. We could really improve readability and maintainability by splitting the component into multiple sub components and better code organization. But good job for now :fire:. Feel free to merge this one.

PR #563, opened 2020-04-27
Refactor plugin route

### LINKED ISSUE #561: Plugin route is not tested
**Please describe the bug:**  
The `/plugin` route is not tested.

1. Run `pytest`
2. See coverage

It should be tested and ideally work with the new API structure.

**Affected Component(s):**  
`CockpitBackend`

### LINKED ISSUE #577: Plugin_settings route is not tested
**Please describe the bug:**  
The `/plugin_settings` route is not tested.

1. Run `pytest`
2. See coverage

It should be tested and ideally work with the new API structure.

**Affected Component(s):**  
`CockpitBackend`

heseding:
# Resolves #561, resolves #577

Refactor the `/plugin` route to be a namespace and test it.

No.

**Affected Component(s):**  
`CockpitBackend`

**Additional context:**  
Adds a `/plugin/<database_id>` route.

**Changes:**  
- [x] Refactor `GET` `/plugin`
- [x] Refactor `POST` `/plugin` (now `POST` `/plugin/<database_id>`)
- [x] Refactor `DELETE` `/plugin` (now `DELETE` `/plugin/<database_id>`)
- [x] Refactor `GET` `/plugin_settings` (to be included in `GET` `/plugin`)
- [x] Refactor `POST` `/plugin_settings` to `PUT` `/plugin/<database_id>`
- [x] Refactor `GET` `/available_plugins` (now `GET` `/plugin/available`)
- [x] Link Plugin Settings to Plugins in the `DatabaseManager`
  _Naming convention for Plugin Settings: `Plugin::{plugin_name}::{setting_name}`, e.g. `Plugin::Compression:MemoryBudget`_
- [x] The `/plugin` namespace is now largely covered by tests
- [x] Example values for all schema fields

**Bonus:**  
- [x] Remove `None` for Plugin Settings
- [x] Remove `Plugin` suffix from plugins, e.g. return `Compression` instead of `CompressionPlugin` (as requested by @MaxSchneider1337)
- [x] Refactor service for `PUT` ~to use a schema~
- [ ] Test `/plugin/available`
- [ ] Test `call_plugin_setting` in `DatabaseManager` (was untested before this PR)


hyrisecockpit/api/app/plugin/controller.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @accepts(schema=PluginSchema, api=api)
+    def post(self, database_id: str) -> Response:
+        """Activate a Plugin in a database."""
+        interface: PluginInterface = request.parsed_obj
+        status = PluginService.activate_by_id(database_id, interface)
+        if status in {200, 404, 423}:
+            return Response(status=status)
+        else:
+            raise ValueError()
```
Alexander-Dubrawski: I think we should handle the errors less generic. In the process of setting the plugin settings, we can get the following status codes: 

internal (inter-process communication) :
* 400: Interprocess communication message couldn't be validated 
* 404: KeyError

external (Request from the client):
* 400: Database not found 
* 423: Database blocked 

We should handle all these cases explicitly. For the internal error status, we could define a custom exception, for all the other unknown exceptions we should define a custom exception that tells us that the status code is unknown.    
Of course, we need to adjust the internal 400 status code.

hyrisecockpit/api/app/plugin/schema.py
```diff
@@ -0,0 +1,57 @@
+"""Schema of a Plugin."""
+from marshmallow import Schema
+from marshmallow.fields import Nested, String
+
+
+class PluginSchema(Schema):
+    """Schema of a Plugin."""
+
+    name = String(description="Identifier of the Plugin.", required=True)
```
Alexander-Dubrawski: I think it would be helpful to add an example description to all schemas: 

```python
name = String(description="Identifier of the Plugin.", example="Compression", required=True)
```

hyrisecockpit/database_manager/manager.py
```diff
@@ -201,10 +206,26 @@ def _call_status(self, body: Body) -> Response:
...
+                for plugin_name in plugins:
+                    if plugin_name in settings.keys():
+                        database_activated_plugins["plugins"][plugin_name] = settings[
+                            plugin_name
+                        ]
+            elif plugins is not None and settings is None:
+                for plugin_name in plugins:
+                    assert database_activated_plugins["plugins"] is not None  # nosec
+                    database_activated_plugins["plugins"][plugin_name] = None
+            activated_plugins.append(database_activated_plugins)
```
Alexander-Dubrawski: I think we can simplify this method. When I read it correctly you are getting the *active* plugins from the database. In the next step, you want to filter the plugin settings for these plugins that are active. My proposal would be to use sets. 

Here is some pseudo-code: 

```python 
plugins = database.get_plugins()
settings = database.get_plugin_setting()

if plugin and setting:
    active_plugins_in_settings = [plugin for plugin in plugins if plugin is in set(settings.keys)]
    return [plugin_setting[plugin] for plugin in  active_plugins_in_settings]
else: 
   return plugins
```

hyrisecockpit/api/app/plugin/service.py
```diff
@@ -0,0 +1,93 @@
...
+                            name=plugin_name,
+                            settings=[
+                                PluginSettingInterface(
+                                    name=plugin_setting["name"],
+                                    value=plugin_setting["value"],
+                                    description=plugin_setting["description"],
+                                )
+                                for plugin_setting in plugin_settings
+                            ]
+                            if plugin_settings is not None
```
Alexander-Dubrawski: Do we need to use `None` values for not existing plug-ins? We could just simply use an empty list instead. Then we wouldn't need to check if the value is `None`. The same goes for `database["plugins"]`.

hyrisecockpit/api/app/plugin/service.py
```diff
@@ -0,0 +1,93 @@
...
+                        )
+                        plugins.append(detailed_plugin)
+                    result.append(
+                        DetailedPluginIDInterface(id=database["id"], plugins=plugins)
+                    )
+                else:
+                    result.append(
+                        DetailedPluginIDInterface(id=database["id"], plugins=None)
+                    )
+            return result
```
Alexander-Dubrawski: Mabey we should move this logic to a separate function. That transforms the response to an object. An alternative would be to just simply hand over the response directly to the controller. The schema can also handle the dictionary (response). I think that approach would be in this case much simpler and easier to read and faster as well (we don't need to cast the response to an object). We would still have the safety of the schema and the type-safety *loss* inside the service is minimal.

Alexander-Dubrawski (commented):
Thanks for the great work so far.

hyrisecockpit/api/app/plugin/schema.py
heseding: Definitely, I put it in the "In Progress" section above and I'll take care of it.

hyrisecockpit/api/app/plugin/service.py
heseding: Returning an empty list of plugins would be interpreted as _"no plugins activated"_. An empty list of settings would, similarly, be interpreted as _"no settings for this plugin"_.
However, we want to communicate that _"**there could be plugins** for this database, but we could not retrieve them (while we could retrieve plugins for the other databases), so here's a partial result"_, and likewise for the settings _"we could not retrieve the settings for this particular plugin, but **there could be settings**"_.
I think it's well worth it.

hyrisecockpit/api/app/plugin/service.py
heseding: > Mabey we should move this logic to a separate function. That transforms the response to an object.

Yes, we could do that.

> An alternative would be to just simply hand over the response directly to the controller.

... which would increase the test effort at the controller level. I think we should clearly separate IPC handling and data transformation into an interface (service) vs. routing (controller).

> The schema can also handle the dictionary (response). I think that approach would be in this case much simpler and easier to read and faster as well (we don't need to cast the response to an object). We would still have the safety of the schema and the type-safety _loss_ inside the service is minimal.

How about this: We use a schema at the service level in order to transform the data to `DetailedPluginIDInterface(many=True)`? This approach would simplify the above method, while we could still separate service and controller responsibilities and keep our type-safety.
Would you be okay with this solution?

hyrisecockpit/api/app/plugin/controller.py
heseding: How would handling these error codes explicitly be more beneficial here, since we're not returning anything except the status code?
I agree that passing the status code blindly can be considered bad practice here, but we're filtering for expected status codes.

Alexander-Dubrawski: Today a got a ValueError exception and it was a problem in the server. I think it would be much nicer to get a custom exception instead of a ValueError. The debugging would also be much easier. 
My argument was to not just throw the internal status codes in the ValueError exception but instead define custom exceptions.

hyrisecockpit/api/app/plugin/service.py
Alexander-Dubrawski: Nice idea 🙃

hyrisecockpit/api/app/plugin/service.py
Alexander-Dubrawski: But if we can't find any settings for a plugin then there are no settings for it? How I understand it we want to return the information which plugins and their settings are loaded and not which could be loaded.

heseding: > But if we can't find any settings for a plugin then there are no settings for it?

I think you're misunderstanding me here.
If there are no settings, `plugin_settings` will be an empty list `[]`.
However, if an error occurs while fetching the settings, `plugin_settings` will be equal to `None` (which means there could or could not be settings for a particular plugin, but we were not able to retrieve that information).

hyrisecockpit/api/app/plugin/service.py
heseding: I put it in the "In Progress" section above and I'll take care of it.

hyrisecockpit/api/app/plugin/service.py
Alexander-Dubrawski: But if there is an error we can't represent a clean status of the plug-ins. In my opinion, we should tell the frontend directly that we aren't able to get the plugins. If we are working with `None` it makes everything more complicated (and the answer we are giving to the frontend is not the real status, because we don't no the real one).  So we should give the frontend all the plugins with a correct status or none at all (by telling the frontend that we are not able to get the information).

hyrisecockpit/api/app/control/app.py
```diff
@@ -79,105 +72,9 @@
...
-            required=False,
-            example="Table not found",
-        ),
-    },
-)
-
 
 @api.route("/available_plugins")
-class ActivatedPlugin(Resource):
+class AvailablePlugin(Resource):
```
PeterTsayun: Would it be possible to move `available_plugins` and `plugin_logs` to the `plugin`-namespace, as they are connected in terms of meaning? Something like `plugin/available_plugins`, `plugin/plugin_logs`?

hyrisecockpit/api/app/plugin/controller.py
```diff
@@ -0,0 +1,69 @@
...
+    def get(self) -> List[DetailedPluginIDInterface]:
+        """Get all Plugins from all databases."""
+        response = PluginService.get_all()
+        if isinstance(response, int):
+            raise ValueError()
+        else:
+            return response
+
+
+@api.response(404, "A database with the given ID does not exist.")
```
PeterTsayun, line 31: Not in this PR, but just like an idea:

The HTTP-error code represents the type of the error, but not the error itself. For example, during execution of the call, the Backend could use 10 resources. If any of them is unreachable, the status code `404: Not found` is returned. 

I think it could be useful when we would return the error message alognside the statuscode from the `DatabaseManager`, because in that case the error occurs in the `DatabaseManager` and manager is responsible for providing of the exact description of the problem, not the Flask application. For example, we could return: 
```
404: The database with the given id does not exist
404: The provided setting name CoolSetting does not exist
404: The provided benchmark name MyBenchmark not found
```
 and so on.  

We could save error message in `response["header"]["error_message"]` and in the Flask application, we would check the status code (`status_code >= 200 or status_code < 300`) and, when needed, set the real HTTP-error-field to the provided error message from the `DatabaseManager`.

hyrisecockpit/api/app/plugin/service.py
PeterTsayun: To my point of view, I am not sure, if returning the "partial correct responses" is a good practice. If I see it correctly, we send only one request `get plugins` to the manager, which is either successful or not. If the request was successful, we have all plugin information about all databases, if not - we have an error message, which is then returned to the user. Is this interpretation correct?

hyrisecockpit/database_manager/manager.py
PeterTsayun: If I see it correctly, `get_plugins` should already return the plugins currently activated in the database. But not all plugins have plugin settings (would be my assumption, please correct me if I'm wrong). So, I think, every activated plugin should be returned anyway, regardless of having corresponding plugin setting or not.

hyrisecockpit/api/app/control/app.py
heseding: Yes, I'll move `available_plugins` to the new namespace, but I suggest we separate `plugin_logs` to #610, does that work for you?

hyrisecockpit/api/app/plugin/controller.py
heseding: Thanks for the suggestion, we should definitely consider it.

> We could save error message in `response["header"]["error_message"]`

How about we save it in `response["header"]["message"]` in order to keep it close to the `Request` interface?

MaxSchneider1337:
could you fix the swagger ui example results. You often Forgot array brackets etc.

Bouncner:
Already a month old, what's the status here?

heseding:
> Already a month old, what's the status here?

It's in review by our backend team.

hyrisecockpit/api/app/plugin/controller.py
```diff
@@ -0,0 +1,76 @@
...
+@api.route("/")
+class PluginController(Resource):
+    """Controller of Plugins."""
+
+    @responds(schema=DetailedPluginIDSchema(many=True), api=api)
+    def get(self) -> List[DetailedPluginID]:
+        """Get all Plugins from all databases."""
+        response = PluginService.get_all()
+        if isinstance(response, int):
+            raise ValueError()
```
Alexander-Dubrawski: I'm still not very happy with the `ValueError`. Why can we not just define a custom Error? Something like:

```python
class IpcException(Exception):
 
    def __init__(self, message: str):
        super().__init__(message)
```

In the code we could use like the following: 

```python
        if isinstance(response, int):
            raise IpcException()
```

hyrisecockpit/database_manager/manager.py
```diff
@@ -201,10 +198,26 @@ def _call_status(self, body: Body) -> Response:
...
+                        plugins={
+                            plugin_name: (
+                                settings[plugin_name]
+                                if plugin_name in settings.keys()
+                                else []
+                            )
+                            for plugin_name in plugins
+                        },
+                    )
+                )
```
Alexander-Dubrawski: I think it would be cleaner to have this logic directly in the database object. We could then just simply call one method in the database object to get the information we need. In that case, the manager would have less logic and wouldn't do the work for the database Object. That would mean less correlation between these to modules.

hyrisecockpit/database_manager/database.py
```diff
@@ -188,54 +188,58 @@ def close_worker(self) -> bool:
         """Close worker."""
         return self._worker_pool.close()
 
-    def get_plugins(self) -> Optional[List]:
+    def get_plugins(self) -> Optional[List[str]]:
```
Alexander-Dubrawski: Since we are only getting the plug-ins with their corresponding settings I would suggest having one entry point for that in the database object. Something like:

```python
def get_plugin_information():

     activated_plugins = self.get_plugins()
     plugin_settings = self.get_plugin_settings()
     ... claculate plugins_setting_object ...
```

Moreover do we really need to ask the meta_plugins table for the plug-ins? 
Are we not getting all active Plugins with a call to the `meta_settings` table in `get_plugin_settings`?

Alexander-Dubrawski (commented):
Thanks for the changes so far. But I have some more minor notes.

hyrisecockpit/database_manager/database.py
caterinamandel98: Yes, we do - not every activated plugin has to have settings, so we could lose activated plugins that way.

heseding: I agree with @caterinamandel98 here: A plugin could be active in `meta_plugins` and have no setting (thus no row) in `meta_settings`, therefore both tables are required.

Additionally, I like it better to separate the request to
- `meta_plugins` with `get_plugins` and
- `meta_settings` with `get_plugin_settings`.

We could agree on a wrapper like you suggested.

Alexander-Dubrawski: Ok thank's for the explanation. I would be very happy about the wrapper 🙃 .

hyrisecockpit/database_manager/manager.py
heseding: It has been moved to the `Database` now.

hyrisecockpit/api/app/plugin/controller.py
heseding: We settled on logging the error and returning a `500` to the frontend instead of throwing exceptions in our backend meeting.
`ValueError` has now been removed.

heseding:
> could you fix the swagger ui example results.

@MaxSchneider1337 has been fixed, thanks for pointing it out!

PeterTsayun (approved):

Alexander-Dubrawski (approved):

PR #564, opened 2020-04-27
B/add time precision

### LINKED ISSUE #497: Provide metric requests with time precision
After #484, the backend will be able to provide historical data for metrics in 1s time intervals. When requesting a time interval of a big length (like 1 hour), it would cause a transfer of 3600 points over a network for each metric. As we won't plot all 3600 in the graph anyway, we need to make aggregation over a suitable time interval.

The frontend should be able to provide precision for a time interval, like 1s, 5s, 30s, 1m. The data should be aggregated in the time interval of the provided length.

Alternatives are welcome.

**Affected Component(s):**  
`App`, `InfluxDB`

**Additional context:**  
Related to #496.

PeterTsayun:
# Resolves #497

Now we are able to specialize precision of the historical data. For that reason, the `precision` HTTP-parameter is added to the historical endpoints. It specializes the length of the aggregation time interval (in ns). The values are calculated as an average.

Examples:

GET (1s-precision)
<pre>
startts:     1587997250000000000
endts:       1587997255000000000
precision:            1000000000
</pre>
Response:
```
[
   {
      "id": "hyrise-1",
      "throuhgput": [
         {
            "timestamp": 1587997250000000000,
            "throughput": 0
         },
         {
            "timestamp": 1587997251000000000,
            "throughput": 67
         },
         {
            "timestamp": 1587997252000000000,
            "throughput": 184
         },
         {
            "timestamp": 1587997253000000000,
            "throughput": 116
         },
         {
            "timestamp": 1587997254000000000,
            "throughput": 138
         }
      ]
   }
]
```
GET (5s-precision)
<pre>
startts:     1587997250000000000
endts:       1587997255000000000
precision:            5000000000
</pre>
Response:
```
[
   {
      "id": "hyrise-1",
      "throuhgput": [
         {
            "timestamp": 1587997250000000000,
            "throughput": 101
         }
      ]
   }
]
```
**Affected Component(s):**  
`App`

**Additional context:**


heseding (approved):

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -438,6 +437,91 @@
...
+    metrics: List[str],
+    points: List[Dict],
+) -> List[Dict]:
+    """Fill missing points with zero."""
+    result: List[Dict[str, float]] = []
+    for timestamp in range(startts, endts, step):
+        new_point: Dict = {"timestamp": timestamp}
+        for metric in metrics:
+            new_point[metric] = 0.0
+        for point in points:
```
Alexander-Dubrawski, line 486: If we would have quite a lot of points and timestamp this loop would take quite some time 🙃 .

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -438,6 +437,91 @@
...
+            if point["time"] == timestamp:
+                for metric in metrics:
+                    new_point[metric] = point[metric]
+                break
```
Alexander-Dubrawski: If we would have quite a lot of points and timestamp this loop would take quite some time 🙃 . Mabey we can just start filling the array completely. After that, we could just check the time of the first and last points and cut the array or fill the missing points with 0.0.  Then the runtime should be the length of the interval.

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -438,6 +437,91 @@
...
+        bind_params={"startts": startts, "endts": endts},
+        epoch=True,
+    )
+    return list(points[table, None])
+
+
+def fill_missing_points(
+    startts: int,
+    endts: int,
+    step: int,
```
Alexander-Dubrawski: I think the code is easier to read if we rename `steps` to `precision`.

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -438,6 +437,91 @@
...
+
+
+def get_historical_metric(
+    startts: int, endts: int, precision_ns: int, table_name: str, metrics: List
+) -> List[Dict[str, Union[str, List]]]:
+    """Get historical metric data for all databases."""
+    result: List = []
+    for database in _get_active_databases():
+        metric_points: List[Dict[str, Union[int, float]]] = get_historical_data(
+            startts, endts, precision_ns, table_name, metrics, database
```
Alexander-Dubrawski, line 515: What would happen if we want the points in an interval of one second but with a precision of 10? Are we handling this case somewhere?

hyrisecockpit/api/app/monitor/app.py
Alexander-Dubrawski: One solution could be (this some pseudo code): 

```python
points = []
new_points
points_index = 0
for timestamp in range(startts, endts, step):
    if points[points_index]["time"] == timestamp:
        new_points.append(points[points_index])
        points_index = points_index + 1
    else: 
        new_points.append(0.0)
```

hyrisecockpit/api/app/monitor/app.py
PeterTsayun: Precision is defined as a length of the subinterval of the `[startts, endts]`-interval, over which aggregation is made. So, `[startts, endts]` must have a minimum length of precision. Otherwise, we would be unable to interpret the returned data.

Alexander-Dubrawski (approved):
Thank's for the great work.

PR #578, opened 2020-04-30
f/fix-precision

schTi:
# Resolves #number

Fixes precision handling.

**Affected Component(s):**  
`Frontend`


MaxSchneider1337 (commented):
could you describe why this did not work before?

schTi:
It already worked fine, there was a bug caused by not correctly updating aurora. In this PR I only added some small  but necessary fixes.
The first one ensures, that all metric requests always have a valid time range. There were a race condition when selecting a big time range when the previous aggregation interval is smaller than the smallest which is available in this time range. Because the mretic controller restarts requests on change it would have started requests with e.g. 60 min range and 1s aggreagtion instead of waiting for the automatic precision adaption to 1 min.
The second one ensures, that we cover the total time range when requesting. So when doing the first historic request I added a higher end time. Otherwise there would be a hole between the first and the second request.

hyrisecockpit/frontend/src/components/selection/PrecisionSelection.vue
```diff
@@ -3,7 +3,7 @@
     id="precision-selection"
     v-model="selectedPrecision"
     :items="selectablePrecisions"
-    label="Aggregation intervals"
+    label="Aggregation Interval"
```
MaxSchneider1337, line 6: thanks for changing this

hyrisecockpit/frontend/src/components/selection/RangeSelection.vue
```diff
@@ -3,7 +3,7 @@
     id="historic-range-selection"
     v-model="selectedHistoricRange"
     :items="availableHistoricRanges"
-    label="Request intervals"
+    label="Observed Time Range"
```
MaxSchneider1337, line 6: see above

MaxSchneider1337 (approved):
good job

Bouncner:
> It already worked fine, there was a bug caused by not correctly updating aurora. In this PR I only added some small but necessary fixes.

Is the Docker environment still not used? If it is, how could this happen?

schTi:
We used the aurora instance for our demo. It not worked there because of some caching issues.

Bouncner:
Then the question is why don't you use the Docker setup? Is it still not usable for developing?

I am wondering why we are talking about Docker for month now and nobody is using it and demos are still not running perfectly.

PR #579, opened 2020-04-30
f/remove-unused-code

schTi:
This PR removes old frontend code, which is not used any more, but still needs maintenance. It also moves some stuff for better code organization.  The production build process will be fixed by this change.

**Affected Component(s):**  
`Frontend`


MaxSchneider1337 (approved):
good stuff love what you did here

PR #580, opened 2020-04-30
Remove table name cache

Alexander-Dubrawski:
Co-authored-by: PeterTsayun <40639972+PeterTsayun@users.noreply.github.com>

# Fix table loading

At the moment we can't load the benchmark tables correctly. Moreover, the endpoint `monitor/status` is not consistent. 

To solve this issue we removed the cached tables. From now on we are asking the hyrise directly which tables are loaded. 

**Affected Component(s):**  
`DatabaseManager`, `Database`, `BackgroundJobScheduler`

## Todo

- [x] adjust `_generate_table_loading_queries` by giving it the original table names
- [x] adjust tests


hyrisecockpit/database_manager/database.py
```diff
@@ -153,39 +145,50 @@ def get_worker_pool_status(self) -> str:
...
+    def get_loaded_tables(self) -> List[Dict[str, str]]:
+        """Return already loaded tables."""
+        try:
+            with self._connection_factory.create_cursor() as cur:
+                cur.execute("show tables;", None)
+                rows = cur.fetchall()
+                result = [row[0] for row in rows] if rows else []
+                return result
+        except (DatabaseError, InterfaceError):
+            return []
```
heseding:
```suggestion
        try:
            with self._connection_factory.create_cursor() as cur:
                cur.execute("show tables;", None)
                rows = cur.fetchall()
        except (DatabaseError, InterfaceError):
            return []
        else:
            return [row[0] for row in rows] if rows else []
```

hyrisecockpit/database_manager/background_scheduler.py
```diff
@@ -415,22 +413,32 @@ def _update_storage_data(self) -> None:
...
             )
 
+    def _get_origin_tables_from_workload_type(
+        self, table_names: List[str], workload_type: str
+    ):
+        return [table_name.split(f"_{workload_type}")[0] for table_name in table_names]
+
     def _generate_table_loading_queries(
-        self, table_names: List[str], folder_name: str
+        self, table_names: List[str], workload_type: str
```
heseding: Could this be renamed back to `folder_name`? We adhered to this naming everywhere.
```suggestion
        self, table_names: List[str], folder_name: str
```

caterinamandel98 (approved):

PR #581, opened 2020-04-30
F/574-running-workload-indicator

### LINKED ISSUE #574: running workload indicator
It is not visible in the frontend whether there is a workload running already or not. So if you start a workload and reload the page, you don't see that you started a workload.

If there is a workload started, the workload should be selected, the frequency should update, the start button should be highlighted and it should not be possible to load or delete data (until you stop the workload). 

**Affected Component(s):**  
`Frontend`

kathwill:
# Resolves #574 

It is not visible in the frontend whether there is a workload running already or not. So if you start a workload and reload the page, you don't see that you started a workload.

**Affected Component(s):**  
`Frontend`


hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -165,6 +172,19 @@ function useWorkloadAction(workload: Ref<Workload>): WorkloadAction {
...
+  getWorkloads().then((response: any) => {
+    if (response.data.length > 0) {
+      workload.value = getWorkloadFromTransferred(response.data[0].folder_name);
+      frequency.value = response.data[0].frequency;
+      frequency.value > 0
+        ? (actions.start.active = true)
+        : (actions.pause.active = true);
+    }
+  });
+  eventBus.$on("DATABASE_ADDED", () => {
```
schTi: Why do you stop the workload, when a new database has been added?

hyrisecockpit/frontend/src/components/workload/WorkloadSelector.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-radio-group class="mt-0">
+  <v-radio-group class="mt-0" :value="workload">
```
schTi: Is this really working? As I see it you cannot change the value of props directly, so you should use the same workaround  used in `FrequencyHandler`

kathwill: This is working, but I can still change it

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: Otherwise you could still have an instance with a running workload and the new one where the right table might not be loaded yet. The alternative would be to load the right workload, stop the workers and start them again, but that can take some time.

schTi: I dont know if this should be a frontend task, you cannot really ensure that you stop the workload correctly. E.g. when reloading the page directly after adding a database the workload will not have been stopped, and the problem persists.

schTi (approved):
The implementation seems to be fine! One thing I noticed: We always have to open the workload generation dialog to see if there is a running workload. Maybe thats not the best solution, and we should display a global message, too. But we can discuss this later on.

PR #589, opened 2020-05-05
B/Finish refactor background scheduler

Alexander-Dubrawski:
This Pr finishes the `BackgroundJobScheduler` refactoring

**Affected Component(s):**  
`BackgroundJobScheduler`, `Database`.

## Todo

- [x] Adjust tests
- [x] ~Move Plugin jobs from `Database` to `Job` and register at `BackgroundJobScheduler`~ Move to seperate PR 
- [x] Add tests for `activate_plugin` job
- [x] Add tests for `dectivate_plugin` job
- [x] Add tests for `delete tables` job
- [x] Add tests for `load tables` job
- [x] Add tests for `execute_queries_parallel` job
- [x] Add tests for `get_loaded_tables` job


hyrisecockpit/database_manager/job/get_loaded_tables.py
```diff
@@ -0,0 +1,24 @@
...
+        with connection_factory.create_cursor() as cur:
+            cur.execute("show tables;", None)
+            results = cur.fetchall()
+            all_loaded_tables = [row[0] for row in results] if results else []
+            for table in tables:
+                if f"{table}_{benchmark}_{scale_factor}" in all_loaded_tables:
+                    loaded_tables.append(table)
+
+    except (DatabaseError, InterfaceError):
+        return []
```
heseding, line 22: How about returning `None` instead of an empty list if an error occurs? An empty list would mean that the query returned no tuples (thus no tables are loaded).

```suggestion
) -> Optional[List[str]]:
    """Check which tables exists and which not."""
    loaded_tables: List = []
    try:
        with connection_factory.create_cursor() as cur:
            cur.execute("show tables;", None)
            results = cur.fetchall()
            all_loaded_tables = [row[0] for row in results] if results else None
            for table in tables:
                if f"{table}_{benchmark}_{scale_factor}" in all_loaded_tables:
                    loaded_tables.append(table)

    except (DatabaseError, InterfaceError):
        return None
```

heseding (approved):

hyrisecockpit/database_manager/job/get_loaded_tables.py
Alexander-Dubrawski: I agree that a return of None would be cleaner. But if we do that every component the uses the `get_loaded_tables_for_scale_factor ` function needs to handle the case that we can't access the hyrise because of its malfunctioning (this would be of cause totally our fault because it is theoretically impossible that a hyrise would crash 👀 ). For the moment we are using the low levelest hyrise error handling as possible (so handle it as soon as possible). So we are just saying that if the hyrise is crashed that there are no tables.

caterinamandel98: I see your point Alex. However, I rather agree with Fabi as we also deal with this in other parts of the code (for example when we request the plugin settings) and I think that the effort is worth it.

Alexander-Dubrawski: Ok, but I would suggest doing it in a separate PR. Is that ok with you?

PR #601, opened 2020-05-07
Adjust system tests

PeterTsayun:
# Resolves #number

Adjust system tests.

No feature added.

**Affected Component(s):**  
`System tests`

**Additional context:**


system_tests/test_system.py
```diff
@@ -160,7 +160,7 @@ def test_execute_workload(self):
         response = self.backend.start_workers()
         assert response.status_code == 200  # nosec
 
-        sleep(5.0)  # wait for query executions
+        sleep(6.0)  # wait for query executions
```
Alexander-Dubrawski: Why did you need to change the sleep time?. This seems to be quite the magic number. That could result in some failings on one system while another would pass the system tests.

PeterTsayun: This sleep interval represents time interval from the starting of the workload to seeing the first metric results. Before the BackgroundScheduler refactoring pull request, 5 seconds were always sufficient. Now, for some reason, the queue length data comes only after 6s. Unfortunately, I couldn't find the problem in the background scheduler so far.

The test was designed, that 5s (which is the upper bound) should be sufficient for any system. If it would not be the case, the test would fail, as expected.

Bouncner: 5s for a test is an enourmous amount of time. We should put some effort into finding the cause.

Which steps are done in this 5s? Have print debugged it to see what part take how much time?

PeterTsayun: 1. The workers are started (for every worker a Python process is created)
2. The Workload Generator is started (The corresponding query files are loaded)
3. Workload Generator publishes a workload to the Queue Worker
4. The Queue Worker pushes the received queries into the workload queue
5. The Workload Workers take queries from the queue and execute them
6. Every second, every worker logs executed query information to Influx
7. We create a request to Backend to get metric information (via Influx)

The problem now is, that after 5s from step 2 we have not always entries in Influx (step 7). 

I will take a closer look at that and do exact measurements.

Bouncner: Why do we put queries into a queue and then in another queue?

PeterTsayun: There is only one workload queue. The Queue Worker process receives the queries via ZMQ sockets (PUB/SUB) from Workload Generator and is responsible for putting them into the workload queue. The workers can then take queries from the queue and execute them.

PeterTsayun: The problem was, that the data was queried for the time interval, which was rounded off in backend resulting in the interval where the requested data didn't exist.

I got the following measurements:
- Starting of the workers - 350ms (100ms for creating the corresponding `BackgroundScheduler` job + 250ms for creating of 8 worker processes)
- Collecting the data about executed queries by workers - 1000ms (we log it every second)
- logging data to Influx - 20ms
- Continuous query calculation for throughput and latency - 1000ms (worst case, Influx makes computation every whole second)
- We use an offset of 1000ms for querying the metric information in the test for rare cases when we request data while the data for the interval is being computed by Influx and thus not available yet.  It happens because the time interval limits are rounded off in the backend. 

So, in total, we have about 3370ms. Now, the requested interval is adjusted and we use 4s sleep time in the test.

Alexander-Dubrawski: Wouldn't it be cleaner to test all these steps in separate tests? So if for example, the workers are not starting we see directly where we have a problem. All the steps could still build on top of each other. 

So we would check if the workers started. Then if the workload generator is running and so on ...

Bouncner: I think Alex is making a good point there with testing each step. Assuming they are "easily" observable.

> * Starting of the workers - 350ms (100ms for creating the corresponding `BackgroundScheduler` job + 250ms for creating of 8 worker processes)

Do you think these numbers sound reasonable? Is there already a lot of data loaded? Jobs and process spawning should at most take single digit milliseconds.

PeterTsayun: > Jobs and process spawning should at most take single digit milliseconds.


I measured it one more time today in the evening (around 9 p. m.) and got the following numbers:

- job creation - 1 ms
- start of the workers - 29 ms
- the time interval from the moment of the process start till the process runs on the CPU: 16 ms

which makes a huge difference to the values that were measured on midday. Sounds like it depends on the number of processes currently running on the virtual machine.



> Assuming they are "easily" observable.

The problem with the separate tests could be, that they need to be isolated from each other. Currently, for that reason, the Backend (Manager + Generator + API) is started and shutdown before and after each test case. Because Flask needs some time for initialization, we wait after starting the Backend for 3s before executing the test (which, again, is a surprisingly big amount of time). If we wait less than 3 seconds `requests` package fails to establish connection to the Flask application. 

So, testing every small action in the separate test case instead of testing use case scenarios in the system tests could increase the execution time significantly.

Alexander-Dubrawski: I am not sure that the tests need to be isolated from each other. What exactly is the benefit? At the moment we are doing some redundant work. For example, we have one test that is adding a database. Another test is doing exactly the same but also starting the worker. 
I think it would be much cleaner if the test would build on each other and using the same backend instance.

For example, we could define the following tests that would build on each other and represent a user Szenario: 

1. start components
2. load tables 
3. start workload generation 
4. start worker 
5. check influxdb tables  
6. check endpoints 
7. add plugin 
8. stop workload generation
9. remove tables 
9. remove database 
10. close components 

Moreover, I would think it would be to use a busy-wait instead of a sleep.  We could still define a time how long the request should take, but also measure the time it took and print it.

Bouncner: > [...] but also measure the time it took and print it.

You could also EXPECT (or whatever that is in your test framework) the time span to be lower to some value.

system_tests/cockpit_backend.py
```diff
@@ -51,22 +51,22 @@ def get_stdout(self):
...
         return output
 
-    def get_monitor_property(self, property: str):
-        """Get monitor property."""
-        url = f"http://{self._backend_host}:{self._backend_port}/monitor/{property}"
-        return get(url, timeout=REQUEST_TIMEOUT).json()
+    def get_property(self, property: str):
+        """Get property."""
+        url = f"http://{self._backend_host}:{self._backend_port}/{property}"
+        return get(url, timeout=REQUEST_TIMEOUT)
```
Alexander-Dubrawski, line 57: Is it possible to print the request time (for all the request methods)? 

For example: 

```python
response = requests.post(url, data=post_fields, timeout=timeout)
print(response.elapsed.total_seconds())
```

system_tests/test_system.py
```diff
@@ -15,44 +15,50 @@
...
-)
 SETUP_TIMEOUT = 3.0
 
 
 class TestSystem:
     """Tests for the whole cockpit system."""
 
-    def setup_method(self, test_method):
+    @classmethod
+    def setup_class(cls):
```
Alexander-Dubrawski, line 25: We could simply name the method `setup` 🙃 .

system_tests/test_system.py
```diff
@@ -15,44 +15,50 @@
...
+        cls.backend.start()
+        cls.manager = CockpitManager()
+        cls.manager.start()
+        cls.generator = CockpitGenerator()
+        cls.generator.start()
         sleep(SETUP_TIMEOUT)
 
-    def teardown_method(self, test_method):
+    @classmethod
+    def teardown_class(cls):
```
Alexander-Dubrawski, line 40: We could name the method `teardown`

system_tests/test_system.py
```diff
@@ -72,56 +78,85 @@ def check_loading_default_tables(self, database_id: str) -> None:
...
         }
 
-        status = self.backend.get_monitor_property("status")
-        assert expected_status in status  # nosec
+        response = cls.backend.get_property("monitor/status")  # type: ignore
+
+        assert response.status_code == 200  # nosec
+        assert expected_status in response.json()  # nosec
+
+    def test_initializes_backend_without_errors(self):
```
Alexander-Dubrawski: I think a better test description would be `test_initializes_backend`.

system_tests/test_system.py
```diff
@@ -72,56 +78,85 @@ def check_loading_default_tables(self, database_id: str) -> None:
...
-        )
-        available_databases = self.backend.get_control_property("database")
-
-        historical_metrics = ["throughput", "latency", "queue_length", "system"]
+            response = self.backend.get_property(metrics[i])
+
+            assert response.status_code == 200  # nosec
+            assert response.json()["body"][metrics_attributes[i]] == {}  # nosec
+
+    def test_returns_historical_metric_values(self):
```
Alexander-Dubrawski: I would rename the test to `test_returns_historical_metric_values_with_no_database_registered`

system_tests/test_system.py
```diff
@@ -72,56 +78,85 @@ def check_loading_default_tables(self, database_id: str) -> None:
...
+        """Test available datasets."""
+        response = self.backend.get_control_property("database/benchmark_tables")
+
+        assert response.status_code == 200  # nosec
+        assert response.json() == {  # nosec
             "folder_names": ["tpch_0.1", "tpch_1", "tpcds_1", "job"]
         }
-        assert available_databases == []  # nosec
 
+    def test_initialization_calls_without_errors(self):
```
Alexander-Dubrawski:
```suggestion
    def test_initialization_calls(self):
```

system_tests/test_system.py
```diff
@@ -131,52 +166,156 @@ def test_database_handling(self):
             }
         ]
 
+    def test_loads_default_tables(self):
+        """Test loading default tables."""
         sleep(5.0)  # wait until default tables are loaded
```
Alexander-Dubrawski, line 171: Mabey we can check with a busy-wait when the tables will be loaded and print it (We could also ask the hyrise directly for that and then check the endpoint). We could still let the test fail if it would take more than 5 seconds.

system_tests/test_system.py
```diff
@@ -131,52 +166,156 @@ def test_database_handling(self):
...
-        influx_databases = influx_client.get_list_database()
+    def test_creates_influx_database(self):
+        """Test the corresponding influx database is created."""
+        influx_databases = self.influx_client.get_list_database()
         assert {"name": "test_database1"} in influx_databases  # nosec
 
-        response = self.backend.remove_database("test_database1")
-        assert response.status_code == 200  # nosec
-
+    def test_handles_database_without_errors(self):
```
Alexander-Dubrawski, line 180:
```suggestion
    def test_handles_database(self):
```

system_tests/test_system.py
```diff
@@ -131,52 +166,156 @@ def test_database_handling(self):
...
 
+    def test_starts_worker_pool(self):
+        """Test starting of the worker pool."""
         response = self.backend.start_workers()
         assert response.status_code == 200  # nosec
 
-        sleep(5.0)  # wait for query executions
+    def test_returns_historical_metric_values_during_workload_execution(self):
+        """Test responses of the historical metrics."""
+        sleep(4.0)  # wait for query executions
```
Alexander-Dubrawski, line 196: We could use here a busy-wait as well.

system_tests/test_system.py
```diff
@@ -131,52 +166,156 @@ def test_database_handling(self):
...
+    def test_activates_plugin(self):
+        """Test activation of the plugin."""
+        response = self.backend.activate_plugin("test_database1", "Compression")
+
+        assert response.status_code == 200  # nosec
+        assert response.json()["header"]["status"] == 200  # nosec
+
+    def test_returns_activated_plugins(self):
+        """Test activation of the plugin."""
+        sleep(1.0)
```
Alexander-Dubrawski, line 221: We could use here a busy-wait as well.

system_tests/cockpit_backend.py
PeterTsayun: What would be the use case of printing it? I think, if we want to test the responsiveness of the endpoints, we could use the corresponding tool with the WRK package, which provides more precise values than `requests`-package.

system_tests/test_system.py
PeterTsayun: The `setup` and `teardown` methods are defined in the `unittest`-module and are executed before and after each test respectively. I think, in this scenario, we want to start the backend components only once before executing all tests and then shutdown them. The corresponding methods from `pytest`-module are `setup_class` and `teardown_class`

system_tests/test_system.py
PeterTsayun: I think, because of using busy-wait, we will not get precise time values anyway. So, what would be the use case of printing them?

system_tests/cockpit_backend.py
Alexander-Dubrawski: I would be just interested to see the request time. It could pinpoint some potential problems.

system_tests/test_system.py
Alexander-Dubrawski: Oh, I see. Then I will submit a complaint to pytest 🤓 .

system_tests/test_system.py
Alexander-Dubrawski: It would be just a pinpoint.

Alexander-Dubrawski (approved):

PR #604, opened 2020-05-07
F/539-add-equalizer-to-workload-generation

### LINKED ISSUE #539: Add equalizer to workload generation
It is not possible to change the frequency of certain queries of a workload

By adding an equalizer to the workload generation dialog it will now be possible

**Affected Component(s):**  
`Frontend`

kathwill:
# Resolves #539 

By adding the equalizer it is now possible to change the weights for the queries of a workload.

**Affected Component(s):**  
`Frontend`

**Additional context:**
The changed workload generation dialog including a tab for the equalizer.
Equalizer12

The equalizer for the started workload with non linear sliders:
Equalizer17

When no workload is started, the equalizer is not enabled (the tip is shown when the mouse is on the equalizer tab):
Equalizer14


hyrisecockpit/frontend/src/components/workload/WeightsHandler.vue
```diff
@@ -0,0 +1,71 @@
+<template>
```
schTi: Thats the equalizer component, so why not call it equalizer?

hyrisecockpit/frontend/src/components/workload/WeightsHandler.vue
```diff
@@ -0,0 +1,71 @@
...
+  <span>
+    <p class="subtitle-1 font-weight-medium">
+      {{ getDisplayedWorkload(workload) }}
+    </p>
+    <v-row>
+      <v-slider
+        v-model="weights[key]"
+        v-for="(weight, key) in weights"
```
schTi: You should change the order of `v-model` and `v-for`. Then you may notice that `weights[key]` is basically `weight`

hyrisecockpit/frontend/src/components/workload/WeightsHandler.vue
```diff
@@ -0,0 +1,71 @@
...
+        :key="key"
+        :label="key"
+        :value="weight"
+        thumb-label
+        thumb-size="20"
+        min="0"
+        max="200"
```
schTi: Is the max value always 200? Can you tell me what unit those weights are measured in? Maybe this would be a useful information for the user, too

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -1,56 +1,111 @@
...
+                          </div>
+                        </template>
+                        <span>
+                          {{
+                            enableEqualizer
+                              ? "Customize workload"
+                              : "Start a workload first"
+                          }}
+                        </span>
+                      </v-tooltip>
```
schTi: In my opinion its enough to switch to the equalizer page via tab header. So just get rid of this button.

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -1,56 +1,111 @@
...
+                    :disabled="runningWorkload || disabled"
+                    @change="handleWorkloadDataChange"
+                  />
+                </v-col>
+              </v-row>
+            </v-card-text>
+          </v-card>
+        </v-tab-item>
+        <v-tab-item>
+          <v-card flat>
```
schTi: Why do you need this card?

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -80,6 +136,7 @@ interface Props {
 
 interface Data extends WorkloadHandler, WorkloadAction, WorkloadDataHandler {
   databases: Ref<readonly string[]>;
+  tab: number;
```
schTi: Should be `Ref<number>`

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -119,12 +180,13 @@ export default defineComponent({
     },
   },
   setup(props: {}, context: SetupContext): Data {
+    const tab: number = 0;
```
schTi: This explicit `number` declaration is unnecessary

schTi (changes requested):

hyrisecockpit/frontend/src/components/workload/WeightsHandler.vue
MaxSchneider1337: this does not work: `'v-model' directives cannot update the iteration variable 'weight' itself.`

schTi: Yes nvm you're absolutely right. But the order can still be changed

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: Otherwise the text color changes, but alternatively I can also add the text color to every text.

hyrisecockpit/frontend/src/components/workload/WeightsHandler.vue
kathwill: There is no real max value. If query 1 has the weight 200, query 2 the weight 100 and there are only these two queries in the workload, 2/3 of all queries are query 1 and 1/3 is query 2. We could think of a relative representation of the weights too.

Bouncner:
> There is no real max value. If query 1 has the weight 200, query 2 the weight 100 and there are only these two queries in the workload, 2/3 of all queries are query 1 and 1/3 is query 2. We could think of a relative representation of the weights too.

I was discussing this with Jan lately. We’ll come back to you.
Do you know if a non-linear increase is possible?

kathwill:
> Do you know if a non-linear increase is possible?

I believe that should be possible.

MaxSchneider1337:
image

Bouncner:
Since this needs to play well with the backend: what is the status there? How would the absolute query counts (as shown in this draft) be used?

kathwill:
> Since this needs to play well with the backend: what is the status there? How would the absolute query counts (as shown in this draft) be used?

Like I said before, so: 
>If query 1 has the weight 200, query 2 the weight 100 and there are only these two queries in the workload, 2/3 of all queries are query 1 and 1/3 is query 2.

So it is for one query always: (weight of query) / (sum of the weights of all queries of this workload)

Bouncner:
If it is a weight, please use 1.0 as the default weight.

But that still doesn't answer how the backend handles these weights.

heseding:
> But that still doesn't answer how the backend handles these weights.

We're using [`random.choices`](https://docs.python.org/3/library/random.html#random.choices) from the Python Standard library [here](https://github.com/hyrise/Cockpit/blob/13dc940d1b4fe8a5a5b4012fda33cf6602aba548/hyrisecockpit/workload_generator/workload.py#L72).
Weights are expected to be `int`, but we can use `float` instead. Initially, all weights are `100`.

The probability of a generating a query is its weight divided by the sum of all weights:
> [If a weights sequence is specified, selections are made according to the relative weights.](https://docs.python.org/3/library/random.html#random.choices)

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,164 @@
...
+          </div>
+        </div>
+        <div class="query-div">
+          <div class="query-name">
+            {{ weight.name }}
+          </div>
+          <div class="top-line"></div>
+          <v-slider
+            v-model="weight.weight"
+            :value="weight.weight"
```
schTi: `:value:` property can be removed here, because this is exactly what `v-model` does

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,164 @@
...
+import { Workload } from "../../types/workloads";
+import { getDisplayedWorkload } from "../../meta/workloads";
+
+interface Props {
+  initialWeights: Object;
+  workload: Workload;
+}
+
+interface Data {
+  weights: Ref<{ name: string; weight: number }[]>;
```
schTi: You can declare a new type here to reuse this declaration. Something like `type Weight = { name: ... }`

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,164 @@
...
+    const weights = ref<{ name: string; weight: number }[]>([]);
+    watch(
+      () => props.initialWeights,
+      () => {
+        weights.value = [];
+        Object.entries(props.initialWeights)
+          .sort()
+          .forEach(([name, weight]: any) => {
+            weights.value.push({ name: name, weight: weight });
+          });
```
schTi: In my optionion you could use `map`  here instead of `forEach`

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -172,14 +209,20 @@ function useWorkloadAction(workload: Ref<Workload>): WorkloadAction {
...
+  let weights = ref<Record<string, number>>({});
 
   getWorkloads().then((response: any) => {
     if (response.data.length > 0) {
       workload.value = getWorkloadFromTransferred(response.data[0].folder_name);
       frequency.value = response.data[0].frequency;
-      frequency.value > 0
-        ? (actions.start.active = true)
-        : (actions.pause.active = true);
+      updatingWorkload().then((response: any) => {
```
schTi, line 223: Can you tell me why you need to call the update workload function here ?

schTi: Is it because you would not get the weights otherwise ?

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
```diff
@@ -172,14 +209,20 @@ function useWorkloadAction(workload: Ref<Workload>): WorkloadAction {
       loading: false,
     },
   });
+  let weights = ref<Record<string, number>>({});
```
schTi: `ref` variables should always be `const`, because you just change the `value` property, but not the declaration itself.

schTi (changes requested):

hyrisecockpit/frontend/src/components/workload/WorkloadGeneration.vue
kathwill: Yes, that's right!

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,193 @@
...
+    },
+    workload: {
+      type: String,
+      default: "",
+    },
+  },
+  setup(props: Props, context: SetupContext): Data {
+    const weights = ref<Weight[]>([]);
+    const a = 50 / 49;
+    const b = Math.pow(99, 1 / 50);
```
schTi: Can you please add some commentary here? Looks like there are some magic numbers and I dont understand where they come from.

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,193 @@
...
+      )!;
+      weight.value = convertSliderValueToValue(sliderValue);
+      context.emit("change", name, weight.value);
+    }
+    function updateTextfield(name: string, value: number): void {
+      const weight: Weight = Object.values(weights.value).find(
+        (weight) => weight.name === name
+      )!;
+      weight.sliderValue = convertValueToSliderValue(value);
+      context.emit("change", name, value);
```
schTi: That is basically the same function except the conversion part. You could merge this to one single function accepting the conversion function as parameter.

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -0,0 +1,193 @@
...
+    const weights = ref<Weight[]>([]);
+    const a = 50 / 49;
+    const b = Math.pow(99, 1 / 50);
+
+    function convertSliderValueToValue(sliderValue: number): number {
+      return Math.round(a * Math.pow(b, sliderValue) - a);
+    }
+    function convertValueToSliderValue(value: number): number {
+      return Math.round(Math.log((value + a) / a) / Math.log(b));
+    }
```
schTi: Can you tell me in a few words, what you need to convert here? Just for my understanding :smile:

schTi (changes requested):
Looks good so far. Just some minor improvement suggestions.

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
kathwill: A non linear increase of the weight values was required, but the v-slider only offers linear increases. So I created an exponential function that converts slider values to our weight values and vice versa. By changing a and b we could still change the value range.

Bouncner:
18 days old. What's the status?

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -94,19 +96,14 @@ export default defineComponent({
...
+    function updateWeight(name: string, sliderValueToValue: boolean): void {
       const weight: Weight = Object.values(weights.value).find(
         (weight) => weight.name === name
       )!;
-      weight.sliderValue = convertValueToSliderValue(value);
-      context.emit("change", name, value);
+      sliderValueToValue
+        ? (weight.value = convertSliderValueToValue(weight.sliderValue))
+        : (weight.sliderValue = convertValueToSliderValue(weight.value));
+      context.emit("change", name, weight.value);
```
schTi: In general its cool to use `?` shortcut for assignments, but it can be really misleading using this with expressions. Maybe good old if and else is more suitable here.

hyrisecockpit/frontend/src/components/workload/Equalizer.vue
```diff
@@ -85,6 +84,9 @@ export default defineComponent({
   },
   setup(props: Props, context: SetupContext): Data {
     const weights = ref<Weight[]>([]);
+
+    /* convert the linear sliderValues with exponential function: f(sliderValue) = value = a * b^sliderValue - a
+    f(0) = 0, f(50) = 100, f(100) =  10000 --> b = 99^(1/50), a = 50/49 */
```
schTi: thanks for the comment. We should always be aware of adding comments to critical functions/ logic. So it would be nice if you can add some comments to all logic, which you thinks needs some further explanation, in the future.

schTi (changes requested):
just one minor thing

heseding:
> If it is a weight, please use 1.0 as the default weight.

Weights are now `float` with `1.0` as default in the backend.

schTi (approved):
Really good work. Merge if you are happy :fire:

Bouncner:
Niiiice!

Looking forward seeing it live in action. 🙂

PR #615, opened 2020-05-12
include log level

### LINKED ISSUE #610: Add log levels to the plugin log endpoint
Related to #605.

1. Requesting the plugin log at `/plugin_log` should pass only the most important warnings by default. Optionally, a parameter `min_log_level` could be passed (e.g. "show all log messages with level 1 and above").
2. The response should include the log level for each log message in plain text (e.g. `Debug`, `Info`, ...).

- Marking the parameter `min_log_level` required, thus breaking the the current interface?
- Adding a parameter `max_log_level`?
- Just using a `log_level` parameter that returns only the specified log level messages?

**Affected Component(s):**  
`CockpitBackend`, `DatabaseManager`

**Additional context:**  
- The endpoint should be ported to the new API structure and tests should be added/adapted.
- Log levels are defined [in Hyrise](https://github.com/hyrise/hyrise/blob/0be3ad615e7b929cd67c5757ce979611bae74174/src/lib/types.hpp#L234).

caterinamandel98:
# Resolves #610 

As plugin events should be shown in the performance graphs and we don't want the graphs to be completely filled with these events, there have to be levels for the log messages. We now provide these for the frontend.

**Affected Component(s):**  
`DatabaseManager`, `App`, ...


heseding (approved):
Looks good to me, I do think however that we should refactor the `/plugin_log` endpoint with the new API structure in mind.
Additionally, I think it would benefit the frontend if we add a parameter to filter by log level.

Alexander-Dubrawski (approved):

MaxSchneider1337:
@caterinamandel98 how does this change the endpoint? Could we get an example request?

PR #620, opened 2020-05-13
f/598-fix-overlapping-data

### LINKED ISSUE #598: Linechart showing incorrect data points and changing existing points
**Please describe the bug:**  
The Linecharts sometimes show incorrect datapoints that jump around x axis values (timestamps should be linear)

1. Go to comparison page
2. see odd stuff

data that is already in the linechart should stay the same. The datapoints should not jump back and forward on the time axis.

**Screenshot(s):**  
image

schTi:
# Resolves #598

This PR ensures, that request intervalls are not overlapping, so that the requested data is not duplicated, which caused incorrect data points.

**Affected Component(s):**  
`Frontend`


MaxSchneider1337 (approved):
looks good

PR #626, opened 2020-05-14
B/adjust system data

### LINKED ISSUE #611: Change system data endpoint
New fields have been added to the [`MetaSystemInformationTable`](https://github.com/hyrise/hyrise/blob/0be3ad615e7b929cd67c5757ce979611bae74174/src/lib/utils/meta_tables/meta_system_information_table.cpp#L16) and [`MetaSystemUtilizationTable`](https://github.com/hyrise/hyrise/blob/0be3ad615e7b929cd67c5757ce979611bae74174/src/lib/utils/meta_tables/meta_system_utilization_table.cpp#L13), thus we do not need to mock some values anymore. Some names have been changed.
Additionally, CPU time is now the total time all CPUs consumed. A relative time can be obtained by dividing the total time with `cpu_count`.

Return the new fields in the backend endpoint.

None.

**Affected Component(s):**  
`DatabaseManager`, `CockpitBackend`

caterinamandel98:
# Resolves #611 

Due to changes of the meta tables in hyrise, we also have to adjust how we request these. Furthermore, we don't have to mock various values anymore since we have real data.

**Affected Component(s):**  
`DatabaseManager`, `App`, ...


heseding (approved):

Alexander-Dubrawski (approved):

hyrisecockpit/api/app/monitor/app.py
```diff
@@ -768,14 +761,13 @@ def get(self) -> Union[int, List]:
...
-                                "cpu_clock_speed": point["cpu_clock_speed"],
                             },
                             "memory": {
                                 "free": point["free_memory"],
-                                "used": point["used_memory"],
+                                "available": point["available_memory"],
                                 "total": point["total_memory"],
                                 "percent": (
-                                    point["used_memory"] / point["total_memory"]
+                                    point["available_memory"] / point["total_memory"]
```
PeterTsayun, line 770: If I see it correctly, "available_memory" is the amount of free memory of the system. I think, the "percent", which represents the used memory,  should be calculated as `"total_memory" - "available_memory" / "total_memory"`

hyrisecockpit/database_manager/job/update_system_data.py
```diff
@@ -7,18 +7,17 @@
...
-    utilization_df, system_df, database_threads: int
-) -> Dict[str, Union[int, float]]:
+def _create_system_data_dict(utilization_df, system_df) -> Dict[str, Union[int, float]]:
     return {
-        "cpu_system_usage": float(utilization_df["cpu_system_usage"][0]),
-        "cpu_process_usage": float(utilization_df["cpu_process_usage"][0]),
+        "cpu_system_usage": float(utilization_df["cpu_system_time"][0])
+        / int(system_df["cpu_count"][0]),
+        "cpu_process_usage": float(utilization_df["cpu_process_time"][0])
+        / int(system_df["cpu_count"][0]),
```
PeterTsayun, line 15: Which interpretation has the `"cpu_system_time"/ "cpu_process_time"` in this case and what is its range? Is that the sum of the percents of the system times of the cores,  or system time in nanseconds per second per core?

caterinamandel98: Beforehand, these columns returned the cpu time per cpu. Now, they return the cumulated values for all cpus. For that reason, we just have to divide the current values by the number of cpus in order to get what we need.

hyrisecockpit/api/app/monitor/app.py
caterinamandel98: Actually, percent doesn't represent the used memory anymore, as we don't get the used memory from the database anymore. Instead, we now get the available memory and based on that, I think it is more useful to use that.

PeterTsayun: As far as I know, the frontend uses the "percent"-field for plotting the used memory data in the RAM graph, so, I would suggest keeping the interpretation of the returned data the same. Moreover, the Unix htop/Windows Process Explorer usually return the used RAM memory in the percents, not the available memory.

PR #629, opened 2020-05-14
f/fix-ram-percent

schTi:
# Resolves #number

The system data endpoint now returns the available memory space instead of the used space in percent. This PR will convert those data correctly.

**Affected Component(s):**  
`Frontend`


MaxSchneider1337 (approved):
good job

MaxSchneider1337:
shouldn't the pr be name fix ram percent?

schTi:
> shouldn't the pr be name fix ram percent?

Yes thats true :smile:

PR #638, opened 2020-05-18
f/637 fix comparison screen

### LINKED ISSUE #637: new comparison design crashes with resolutions below 4k
**Please describe the bug:**  
the new comparison screen makes it impossible to read numbers, when more then 1 db is selected at resolutions below 4k.

1. comparison screen
2. See error

the text should be readable on all resolution >=1080p

**Screenshot(s):**  
image

**Affected Component(s):**  
Frontend

MaxSchneider1337:
# Resolves #637

fixes the comparison screen design, because it only works on xl resolutions

**Affected Component(s):**  
Frontend

**Additional context:**  
image


MaxSchneider1337:
without some of the whitespace
image
image

Bouncner:
Shouldn't that safely fit into two rows?

MaxSchneider1337:
> Shouldn't that safely fit into two rows?
 
2 rows does not work with my resolution, since Data Size and Memory Capacity Strings can get pretty long.

3 selected Databases would also destroy the formatting in that case (edited)

Bensk1:
> 2 rows does not work with my resolution, since Data Size and Memory Capacity Strings can get pretty long.

We could display the data size in GB, that should save a bit of space. If necessary, we could also go for `# of CPUs`?

MaxSchneider1337:
> Shouldn't that safely fit into two rows?

The comparison-screen only looks like this on resolutions below 1904px

Bouncner:
So it looks like it on the most popular MacBook? ("13-inch MacBook Pro models introduced in late 2012 or later. Native resolution: 2560 x 1600 at 227 pixels per inch. Support for millions of colors.")

MaxSchneider1337:
> So it looks like it on the most popular MacBook? ("13-inch MacBook Pro models introduced in late 2012 or later. Native resolution: 2560 x 1600 at 227 pixels per inch. Support for millions of colors.")

Yes and we are using every single color ;-)

I will try a different layout and let you now once it is ready

MaxSchneider1337:
@Bouncner what do you think about this design: 
image
image

MaxSchneider1337:
> @Bouncner what do you think about this design:
> image
> image

we could also

Bouncner:
If it takes up that much space, we should redesign it or make is "minimizable" IMO. That's just too much space wasted for static information.

Bouncner:
Also, can we group the metric correctly? Host name, host stats (i.e., capacity and cpus), footprint, and workers.

MaxSchneider1337:
> Also, can we group the metric correctly? Host name, host stats (i.e., capacity and cpus), footprint, and workers.

yes of cause

hyrisecockpit/frontend/src/components/details/DatabaseSystemDetails.vue
```diff
@@ -6,48 +6,84 @@
...
-        <v-row>
-          <v-col id="database-host" class="leftcol" cols="3">
-            <v-card-title class="ml-6 font-weight-regular white--text">
+      <v-card class="card" :color="database.color">
+        <v-row class="detailsRow">
+          <v-col
+            id="database-host"
+            :lg="databases.length > 1 ? 4 : 3"
+            xl="3"
+            v-bind:class="{
```
schTi: `v-bind` prefix can go here, `:` is basically the shortcut for this.

schTi (commented):
Looks great. Is there anything missing you want to add?

MaxSchneider1337:
> Looks great. Is there anything missing you want to add?

I don't really know... The fix is already done so this issue is ready. The best approach to me would be to write a new issue for improving the design, since this would be a pretty big task.

what do you think @Bouncner ?

Bouncner:
Is it possible to scale the font for all characteristics (not the host name) with the available width. The additional lines really bother me. They are additional information and take up by far too much space at the beginning of the page and with their background they absolutely pull the user's attention.

monasobh:
I also think we should make the font smaller. If all the metric numbers still don't fit into one row, I would suggest to leave the host name and - depeding on how many fit - one or two metric numbers in one row and add a button on the side of the page to be able to expand the db information boxes.

MaxSchneider1337:
> Is it possible to scale the font for all characteristics (not the host name) with the available width. The additional lines really bother me. They are additional information and take up by far too much space at the beginning of the page and with their background they absolutely pull the user's attention.

scaling the text does not really work for me, because the text gets really small with 3 dbs.

MaxSchneider1337:
> Is it possible to scale the font for all characteristics (not the host name) with the available width. The additional lines really bother me. They are additional information and take up by far too much space at the beginning of the page and with their background they absolutely pull the user's attention.

image
 this is how it looks with scaled text, really sorry for all the confusion this caused, but I just realized, that the MacBook scales the browsersize to 1440x900 px

Bouncner:
Can we use "Workers" and "CPUs" ... I think the "number of" is useless.

I still think the fonts can be smaller though but that's up to you.

MaxSchneider1337:
image

This is the current version on a small screen

MaxSchneider1337:
> Can we use "Workers" and "CPUs" ... I think the "number of" is useless.
> 
> I still think the fonts can be smaller though but that's up to you.

already fixed this yesterday, forgot to push, sorry

Bouncner:
I know

Bouncner:
Whooopsie. Sorry.

Bouncner:
I know, I am a pain in the ass. But can you try to increase the distinction between the host part on the left and the "remainder". I think we could use smaller fonts for the remainder and please also decrease the vertical space between the lines (Memory Capacity).

MaxSchneider1337:
@Bouncner what do you think about this?

image

Bouncner:
Maybe right align all the "remainders"? But that's up to you. It's just that I think the space smallest space between all the boxes is between the title and the remainder. There should be the biggest gap.

Apart from that: LGTM.

MaxSchneider1337:
image
this is how the screen looks now so the different chart header elements do not overlap on smaller screens. All of the icons show the db name on hover.

MaxSchneider1337:
> image
> this is how the screen looks now so the different chart header elements do not overlap on smaller screens

schTi (approved):
Looks good, feel free to merge if you are happy :fire:

PR #642, opened 2020-05-19
B/status information namespace refactoring

Alexander-Dubrawski:
# Continue monitor namespace refactoring

This pr is creating a new `status` namespace with the following endpoints:

* `/database` returns the *database_blocked_status*, *worker_pool_status* and `hyrise_status`
* `/benchmark` returns the loaded benchmark and benchmark tables
* `/failed_tasks` returns the failed tasks

**Affected Component(s):**  
`DatabaseManager`, `App`

**Additional context:**  
I decided to split the `/status` endpoint in the two `/database` and `/benchmark` endpoints. I think the endpoint was a little bit overloaded. Moreover if the frontend needs to check if the hyrise is alive, we don't need to calculate the loaded benchmarks and tables all the time. 

## Todo

- [x] Write tests for controller
- [x] Write tests for model 
- [x] Write tests for schema 
- [x] Write tests for service 
- [x] Remove empty interface file 
- [x] Remove status and failed task endpoint from the old `app` file 
- [x] Remove status call from Manager and tests for it 
- [x] Adjust system tests
- [x] Move hyrise_stauts to database endpoint


PeterTsayun (commented):
I don't really understand why do we need to split the `/status` endpoint. In the past, we did some measurements, and they showed that the response time of `/status` was about 50ms, which is quite ok. Moreover, I found it more convenient from the architectural point of view to have all information about the current status of the databases in one endpoint rather than calling multiple endpoints to get the information pieces, which are semantically close. It would reduce context switches as well (or however they are called). 
I think we should have discussed that proposal before implementing it.

Alexander-Dubrawski:
> I don't really understand why do we need to split the `/status` endpoint. In the past, we did some measurements, and they showed that the response time of `/status` was about 50ms, which is quite ok. Moreover, I found it more convenient from the architectural point of view to have all information about the current status of the databases in one endpoint rather than calling multiple endpoints to get the information pieces, which are semantically close. It would reduce context switches as well (or however they are called).

I have another opinion than you. I think the endpoint was doing more than one thing:

* Return if the hyrise is active 
* Return if the worker pool is running 
* Return if the database is blocked 
* Return which benchmarks are loaded 
* Return which tables are loaded

In my opinion, to get the hyrise status, the status of the database object and benchmarks are logically different. That was the main motivation to split it and to refactor it. I think the code is now easier to read and the endpoints are cleaner as well. 
Moreover, we are not doing unnecessary work if we just want to know if the hyrise is active. At the moment the frontend is fetching all the time the `/status` endpoint to get the information if the hyrise is alive. In the backend, the following happens: 

* Get hyrise is alive flag 
* Get worker_pool status 
* Send a request to the hyrise which tables are loaded 
* calculate the loaded benchmarks 

For the intention of the request (get the status of the hyrise) we should just get the hyrise alive flag.
So I think from an architectural standpoint of view its just cleaner to do just the work we should do.

> I think we should have discussed that proposal before implementing it.
 
I agree that a discussion is always good, but for me, it was a part of the refactoring process and pretty clear that a split of the endpoint is needed 🙃 .

heseding:
It would be interesting to hear how our frontend team uses the `/status` endpoint and if they would benefit from splitting it into the four endpoints
1) `/status/hyrise`,
2) `/status/database`,
3) `/status/benchmark` and
4) `/status/failed_tasks`.

@schTi @MaxSchneider1337 @kathwill @monasobh We would highly appreciate your input on this.

Alexander-Dubrawski:
> It would be interesting to hear how our frontend team uses the `/status` endpoint and if they would benefit from splitting it into the four endpoints
> 
> 1. `/status/hyrise`,
> 2. `/status/database`,
> 3. `/status/benchmark` and
> 4. `/status/failed_tasks`.
> 
> @schTi @MaxSchneider1337 @kathwill @monasobh We would highly appreciate your input on this.

I talked to the frontend gang. They prefer the following separation of the endpoint:

* `/database` Includes the hyrise status, database_blocked_status, worker_pool_status
* `/benchmark` loaded tables and benchmark 
* `/failed_task_queue`

They are asking for the hyrise_status and database_blocked_status in one request and in another request they are asking for the benchmark information. So a separation makes sense.

hyrisecockpit/api/app/status/service.py
```diff
@@ -0,0 +1,65 @@
...
+        """Get failed task from databases."""
+        results = []
+        with StorageConnection() as client:
+            for database in _get_active_databases():
+                failed_queries = list(
+                    client.query(
+                        "SELECT * FROM failed_queries LIMIT 100;", database=database,
+                    )["failed_queries", None]
+                )
+                serilized_failed_queries = [
```
PeterTsayun:
```suggestion
                serialized_failed_queries = [
```

hyrisecockpit/api/app/status/service.py
```diff
@@ -0,0 +1,65 @@
...
+                    FailedQuery(**query) for query in failed_queries
+                ]
+                results.append(
+                    FailedTask(id=database, failed_queries=serilized_failed_queries)
```
PeterTsayun:
```suggestion
                    FailedTask(id=database, failed_queries=serialized_failed_queries)
```

PeterTsayun (approved):
Thanks for the nice job!

PR #648, opened 2020-05-25
handle monotonically increasing cpu time values

### LINKED ISSUE #633: Unrealistic high cpu usage numbers
**Please describe the bug:**  
We are getting unrealistic high values from the `/system` endpoint.  

1. Add a database.
2. get data from the `/system` endpoint

`cpu_process_usage` and `cpu_system_usage should` be between 0 and 100%.

**Screenshot(s):**  
IMAGE 2020-05-16 11:48:21

**Affected Component(s):**  
`DatabaseManager`

**Additional context:**  
It is interesting to see that the `cpu_process_usage` seems to grow linear even if no workload is running.

IMAGE 2020-05-16 11:49:47

caterinamandel98:
Co-authored-by: Fabian Heseding <fabian.heseding@web.de>

# Resolves #633 

The current cpu usage values are monotonically increasing, unlike the former values, so we have to handle that.


hyrisecockpit/database_manager/job/update_system_data.py
```diff
@@ -49,6 +49,15 @@ def update_system_data(
     time_stamp = time_ns()
 
     with storage_connection_factory.create_cursor() as log:
+        result = list(
+            log._connection.query(
+                "SELECT * FROM system_data ORDER BY time DESC LIMIT 1;",
```
PeterTsayun, line 54: How about storing the previous CPU usage value in a variable, like in `update_chunks_data`? It would reduce the context switches (or however they are called), as access of a local variable in Python is much faster than requesting the data from Influx.

hyrisecockpit/database_manager/job/update_system_data.py
```diff
@@ -49,6 +49,15 @@ def update_system_data(
...
+                database=log._database_id,
+            )["system_data", None]
+        )
+        if result:
+            system_data["cpu_system_usage"] -= result[0]["cpu_system_usage"]
+            system_data["cpu_process_usage"] -= result[0]["cpu_process_usage"]
```
PeterTsayun, line 60: Is that correct, that Hyrise stores the total execution time of ns since the database starting in `cpu_process_usage` and `cpu_system_usage`?
Would it imply, that the very first call typically returns a huge number?

hyrisecockpit/database_manager/job/update_system_data.py
heseding: > access of a local variable in Python is much faster than requesting the data from Influx.

We would have to measure whether that makes much of a difference.
Currently, we want to stay away from storing the same data twice as a means of optimization.

PeterTsayun: I've just done the measurements, and the Influx request take about 32 milliseconds, whereas accessing a Python variable takes just 3 microseconds, which makes the difference of more than 1000 times.

Moreover, I think, that we need to take into the consideration, that Influx is a shared resource between jobs, and in my opinion, we should keep it free if we don't have the direct need to access it.

caterinamandel98: I see your point. However, I think it is cleaner and not that error-prone to obtain these data directly from the influx.

hyrisecockpit/database_manager/job/update_system_data.py
caterinamandel98: Yes.

PeterTsayun: Would it imply, that instead of subtracting the last `cpu_system_usage`/`cpu_process_usage` value from the result, we need to subtract the sum of all previous values? It would explain the alternating entries in the `system_data`-table (b/system_data_fix branch):


```
time                available_memory cpu_count cpu_process_usage cpu_system_usage database_threads free_memory total_memory
----                ---------------- --------- ----------------- ---------------- ---------------- ----------- ------------
1590476191714885126 30679273472      16        782832028.625     49112241250000   16               13258350592 33725243392
1590476192709732611 30677266432      16        99033760.625      59375000         16               13256458240 33725243392
1590476193712637351 30677204992      16        890197960.3125    49112300000000   16               13256286208 33725243392
1590476194758054777 30677508096      16        192836393.6875    111875000        16               13256245248 33725243392
1590476195710818598 30600568832      16        978324696.625     49112588750000   16               13179351040 33725243392
1590476196714856058 29967446016      16        556226314.8125    512500000        16               12506525696 33725243392
1590476197720167052 29823025152      16        1385047164.25     49112976250000   16               12228222976 33725243392
1590476198714432021 29532131328      16        1028165026.375    1006250000       16               11902758912 33725243392
1590476199779380514 29483696128      16        1838657530.625    49113426250000   16               11788681216 33725243392
1590476200716263500 29497356288      16        1479430049.8125   1483125000       16               11788271616 33725243392
1590476201707919424 29503152128      16        2236573233        49113800000000   16               11802365952 33725243392

```

hyrisecockpit/database_manager/job/update_system_data.py
Alexander-Dubrawski: I don't really see why we need to access the influxDB to get the last system data. Like Peter said it would be easier and simpler to simply store it in a global variable. We know that no other job has access to it. @caterinamandel98  why do you think it is less error-prone? I think we have the same consistency if we store it in a variable or in the influxDb.  Both options have the same susceptibility to errors. The influx is not consistent or the variable. Both shouldn't happen, because only the update_system_data job writes in the influx or accessing the variable.

heseding (approved):

Alexander-Dubrawski:
@fabianhe @caterinamandel98  Why did you merge this PR? 
We still have altered values. 

IMAGE 2020-06-02 20:31:20

Bouncner:
Guys, we see merges of branches that shouldn't been merged (yet) pretty much every week. What's up here?

PR #650, opened 2020-05-25
f/645-adjust-query-type-chart

### LINKED ISSUE #645: adjust query type proportion chart
We are now working with real krueger data and the structure of the returned data has changed. Thats how it looks like:

`[
  {
    "id": "momentum",
    "krueger_data": [
      {
        "query_type": "SELECT",
        "total_latency": 18121695583,
        "total_frequency": 831
      },
      {
        "query_type": "CREATE",
        "total_latency": 286013,
        "total_frequency": 37
      },
      {
        "query_type": "UPDATE",
        "total_latency": 0,
        "total_frequency": 0
      },
      {
        "query_type": "INSERT",
        "total_latency": 0,
        "total_frequency": 0
      },
      {
        "query_type": "DELETE",
        "total_latency": 0,
        "total_frequency": 0
      },
      {
        "query_type": "DROP",
        "total_latency": 51837,
        "total_frequency": 37
      },
      {
        "query_type": "COPY",
        "total_latency": 2994107353,
        "total_frequency": 8
      },
      {
        "query_type": "OTHER",
        "total_latency": 0,
        "total_frequency": 0
      }
    ]
  }
]`

The chart logic has to be changed to handle this data change. Additionally latency and frequency should be shown as tooltip information. 

Which data should be used for the visual representation inside the chart, @Bouncner ?

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #645

This PR adjusts the query type chart to the new data structure.

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/src/components/container/MetricsComparisonTable.vue
+++ b/hyrisecockpit/frontend/src/components/container/MetricsComparisonTable.vue
@@ -7,12 +7,16 @@
       :key="database.id"
       :color="database.color"
     >
-      <v-card tile class="column-top-border" :color="database.color">
+      <v-card
+        v-if="showHeader"
+        tile
+        class="column-top-border"
+        :color="database.color"
+        flat
+      >
         <v-card-text class="white--text text-center">
           <strong>
-            {{
-            database.id
-            }}
+            {{ database.id }}
           </strong>
         </v-card-text>
       </v-card>
@@ -29,9 +33,7 @@
       </div>
       <v-card tile class="column-bottom-border" :color="database.color">
         <v-card-text class="white--text text-center">
-          {{
-          database.id
-          }}
+          {{ database.id }}
         </v-card-text>
       </v-card>
     </div>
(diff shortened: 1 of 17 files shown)


schTi:
Here is what it looks like:

image

image

image

@Bouncner @Bensk1 What do you think about this adapted chart?

Bouncner:
1.) "Aggregated runtime" instead of "latency"? It's an aggregation.
2.) It's not "query type" but "operator".
3.) I like the chart very much. I was wondering if we really want the frequency part in the instance page (I know that was my request). But then again I wouldn't know which DB's latency to show on the workload page. Since I see value in the chart, I'd keep it the current way.

schTi:
Just to make sure we are talking about the same thing. This chart is not the chart displaying the real operator data, it will have entries like `SELECT`, `UPDATE`, ... . Maybe it would be a bit misleading to call it operators, too? Additionally this chart is currently shown on the workload analysis page together with the already existing operator runtimes overview barchart. On the instance page only the operator runtimes overview chart will be shown. 

To handle the problem of not knowing which data should be displayed on the workload analysis page I decided to show both charts for every database like you can see in those screenshots.

Bouncner:
Yes, I was mixing up things. Sorry.

In the backend, I think we agreed to use "statements". So operators was wrong, but please change it from query type to "statement type". @Bensk1: am I right?

So 1.) remains, 2.) it's probably "statement", and 3.) I guess I was drunk.

MaxSchneider1337 (approved):
really good changes

PR #594, opened 2020-05-06
B/refactor monitor namespace

PeterTsayun:
# Resolves #576

**Does your pull request solve a problem? Please describe:**  
Refactoring of the monitor namespace.

**Does your pull request add a feature? Please describe:**  
No feature added.

**Affected Component(s):**  
`App`

## Todo

- [x] Add tests for controller
- [x] Add tests for interface
- [x] Add tests for model
- [x] Add tests for schema
- [x] Add tests for service
- [x] Adjsut monitor app
- [ ] Inform Frontend gang about new routes
- [x] Add return types in service 🤓 
- [x] Add models to hand over to service

PR #668, opened 2020-06-09
B/add tpcc

Alexander-Dubrawski:
# Add support for TPCC Benchmark and generic workload drivers

This PR implements the support for the tpcc benchmark. 

## Adjusted architekture

After a first working draft, we realized that our current backend architecture is not well suited to support different types of workloads. In comparison to the tpcds, tpch and job benchmarks the tpcc benchmark decides dynamically on the result of a query what query it should execute next. So we would have different workload/benchmark logics clustered in the whole backend. We decided that it is needed to separate the benchmark/workload logic from the cockpit logic. The cockpit's only task is to execute the task and get all relevant meta information. We created drivers for the workload/benchmarks that define the logic for them.

image

The generator uses the driver to generate the tasks. The database uses the driver to execute the tasks. A driver needs to implement the following interfaces: 

* get_scale_factors() *return the supported scale factors for a workload*
* get_default_weights() *return the default weights for a workload*
* generate(scale_factor, frequency, weights) *return the generated tasks*
* get_table_names(scale_factor) *return the table names of a workload and their representation in the database*
* get_load_queries(scale_factor)*return the table names representation in the database and the SQL querie to load them*
* get_delete_queries(scale_factor)*return the table names representation in the database and the SQL querie to delete them*
* execute_task(task, cursor, worker_id) *executes the task on the hyrise database and returns the metainformation*

We decided that the best generic implementation is to find an abstraction for the workloads. A workload is defined by a name (type) and its scale factor. The Cockpit only understands a workload by these two attributes and it communicates with these two attributes with the driver. How exactly the tables are named in the database and in which folder the queries are located (or how they are generated) is implemented in the driver. As a result of that, we needed to adjust a few endpoints. The frontend can for example load the tables by passing a JSON like `{"workload_type": tpch, "scale_factor": 10.0}` instead of using the folder name of the queries like `tpch_1`. That is especially helpful with benchmarks like the tpcc.  

The drivers and the cockpit are connected via the connector. The connector is managing a dictionary with all drivers for the database component and a dictionary with Workload Objects (that are including the driver) for every workload that the generator can use. 

## TPCC implemeentation 

We implemented the tpcc driver with the help of the https://github.com/odanivan/py-tpcc repository. Since the benchmark is crating conflicts between the transaction it can happen that one worker needs to cancel a transaction. For that, we defined a *negative* throughput. It describes the number of transactions that couldn't be committed because of a Transaction conflict (For more information see https://en.wikipedia.org/wiki/Optimistic_concurrency_control). In the *successful_querie* table (in the influx) we have now a flag that shows if the transaction was committed (successful) or not committed (canceled).

### Database 
* (get) /control/database/benchmark_tables -> (get) /control/database/workload_tables
* (delete) /control/database/benchmark_tables -> (delete) /control/database/workload_tables
* (post) /control/database/benchmark_tables -> (post) /control/database/workload_tables

### Status 
* (get) /status/benchmark -> (get) /status/workload_tables

### Workload 
* (get) /workload/
* (post) /workload/
* (put) /workload/{folder_name} -> merged into (post) /workload/
* (get) /workload/{folder_name} -> (get) /workload/{workload_type}
* (delete) /workload/{folder_name} -> (delete) /workload/{workload_type}

### Adjust 
- [x] Remove running argument from `/workload/{workload_type}`
- [x] Add weights argument to `/workload/{workload_type}`. If weights is empty don't change it in workload
- [x] merge Post and Put in `/workload/{workload_type}`
- [x] return `running, scale_factor, workload_type, frequence, default_scale_factor, weights` in `/workload/`
- [x] return one Object per workload with loaded tables, scale factor and if all tables for it are loaded in `/status/workload`
- [x] get all supported scale factors in `/control/database/workload_tables` (database should get this information from the connector) and return table names for workload
- [x] Adjust storage log to log scale factor
- [ ] Adjust continuous queries to calculate positive and negative throughput
- [x] remove readme from job_metainformation
- [x] Add job and tpcds to connector 
- [x] Remove old duplicated code
- [x] api.database.service.get_available_workload_tables should get the Information from connector
- [x] reference source https://github.com/apavlo/py-tpcc
- [x] fix TPC-C rollback handling

### Implement 
- [x] driver for tpcds 
- [x] driver for job 
- [x] driver for tpcc 
- [ ] move queries to driver (maybe in a seperate PR?)
- [ ] Endpoint for negative throughput (wait for #594) (maybe in a seperate PR?)

### Adjust tests
- [x] worker_pool
- [x] load/delete Table jobs
- [x] background job manager

### Write tests
- [x] default driver 
- [x]  tpcds driver
- [x]  job driver
- [x]  tpcc driver
- [x]  tpch driver
- [x] execute task worker
- [x] workload

### Anything else 
- [x] inform backend mates (apologize for the size of this pr)
- [x] write a good pr description 
- [x] inform frontend about endpoint changes


hyrisecockpit/database_manager/worker/handler/tpcc/tpcc_parameter_generator.py
```diff
@@ -0,0 +1,256 @@
+# noqa
+# type: ignore
+# flake8: noqa
+
+"""Module for TPC-C parameters generation."""
```
cH3n7i: While all/most other files in this tpcc folder don"t need the prefix `tpcc_` this file does?
The files should either all have prefix according to their type of workload or none (and its given away by the folder name)

Bensk1:
Is it possible to "repair" this PR without too much effort? Repairing in the sense of fixes the misleading diffs and decreasing the number of files from 226 to something that makes sense.

If it's too much of a pain, i.e., take more than 30 minutes, we should probably stick with it.

hyrisecockpit/workload_generator/generator.py
```diff
@@ -76,81 +75,94 @@ def _init_server(self) -> None:
     def _call_get_all_workloads(self, body: Body) -> Response:
         response = get_response(200)
         response["body"]["workloads"] = [
-            {"folder_name": folder_name, "frequency": workload.frequency}
-            for folder_name, workload in self._workloads.items()
+            {
+                "workload_type": workload,
+                "frequency": properties.frequency,
+                "scale_factor": properties.scale_factor,
```
PeterTsayun, line 78: Maybe we should return `weights` as well

hyrisecockpit/drivers/default.py
```diff
@@ -0,0 +1,190 @@
...
+                ),
+            )
+        return load_queries
+
+    def get_delete_queries(self, scalefactor):
+        """Generate delete workload queries."""
+        if scalefactor < 1.0:
+            scalefactor = str(scalefactor).replace(".", "_")
+        else:
+            scalefactor = str(int(scalefactor))
```
PeterTsayun: How about
```
if scalefactor == int(scalefactor):
    formatted_scalefactor = str(int(scalefactor))
else:
    formatted_scalefactor = str(scalefactor).replace(".", "_")
```

hyrisecockpit/workload_generator/generator.py
Alexander-Dubrawski: At the moment we are getting detailed information from `_call_get_workload `. We can ask the frontend if it would be helpful for them to get this information from this endpoint es well.

hyrisecockpit/drivers/default.py
Alexander-Dubrawski: What exactly do you mean with `if scalefactor == str(scalefactor)`? Scalefactor is a float, so it would be all the time false.

PeterTsayun: Sorry, I mean `scalefactor == int(scalefactor)`

hyrisecockpit/drivers/job/job_meta_information.py
```diff
@@ -0,0 +1,141 @@
+"""Meta information for job workload."""
+
+default_weights = {
```
heseding, line 3:
```suggestion
default_weights = {key: 1.0 for key in ['10a', '10b', '10c', '11a', '11b', '11c', '11d', '12a', '12b', '12c', '13a', '13b', '13c', '13d', '14a', '14b', '14c', '15a', '15b', '15c', '15d', '16a', '16b', '16c', '16d', '17a', '17b', '17c', '17d', '17e', '17f', '18a', '18b', '18c', '19a', '19b', '19c', '19d', '1a', '1b', '1c', '1d', '20a', '20b', '20c', '21a', '21b', '21c', '22a', '22b', '22c', '22d', '23a', '23b', '23c', '24a', '24b', '25a', '25b', '25c', '26a', '26b', '26c', '27a', '27b', '27c', '28a', '28b', '28c', '29a', '29b', '29c', '2a', '2b', '2c', '2d', '30a', '30b', '30c', '31a', '31b', '31c', '32a', '32b', '33a', '33b', '33c', '3a', '3b', '3c', '4a', '4b', '4c', '5a', '5b', '5c', '6a', '6b', '6c', '6d', '6e', '6f', '7a', '7b', '7c', '8a', '8b', '8c', '8d', '9a', '9b', '9c', '9d']}
```

hyrisecockpit/drivers/tpcds/tpcds_meta_information.py
```diff
@@ -0,0 +1,71 @@
+"""Meta information for tpcds workload."""
+
+default_weights = {
```
heseding, line 3:
```suggestion
default_weights = {key: 1.0 for key in ['01', '03', '06', '07', '09', '10', '13', '15', '17', '19', '25', '26', '28', '29', '31', '34', '35', '39a', '39b', '41', '42', '43', '45', '48', '50', '52', '55', '62', '65', '69', '73', '79', '81', '83', '85', '88', '91', '93', '96', '97', '99']}
```

hyrisecockpit/drivers/tpch/tpch_driver.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def get_scalefactors(self):
+        """Get supported scalefactors."""
+        return self.scale_factors
+
+    def get_default_weights(self):
+        """Get default weights."""
+        return OrderedDict(
+            ("{:02d}".format(i), 1.0) for i in range(1, 23)
+        )  # TODO why OrderedDict
```
heseding, line 45: Why are the default weights not set in a file like `tpch_meta_information` as done for JOB and TPC-DS?

caterinamandel98 (approved):

hyrisecockpit/api/app/database/interface.py
```diff
@@ -19,13 +19,14 @@ class DetailedDatabaseInterface(DatabaseInterface):
...
+    workload_type: str
+    scale_factor: float
 
 
-class AvailableBenchmarkTablesInterface(TypedDict):
+class AvailableWorkloadTablesInterface(TypedDict):
     """Interface of available benchmark tables interface."""
 
-    folder_names: List[str]
+    workload_tables: List
```
heseding, line 32:
```suggestion
    workload_tables: List[WorkloadTablesInterface]
```

hyrisecockpit/api/app/database/service.py
```diff
@@ -62,22 +63,29 @@ def deregister_database(cls, interface: DatabaseInterface) -> int:
...
+        drivers = Connector.get_workload_drivers()
+        workload_tables = []
+        for workload_type, driver in drivers.items():
+            for scale_factor in driver.scale_factors:
+                workload_tables.append(
+                    WorkloadTables(
+                        workload_type=workload_type, scale_factor=scale_factor
+                    )
+                )
+        return AvailableWorkloadTables(workload_tables=workload_tables)
```
heseding: See [PEP 202](https://www.python.org/dev/peps/pep-0202/):
> List comprehensions provide a more concise way to create lists in situations where [...] nested loops would currently be used.

```suggestion
        return AvailableWorkloadTables(
            workload_tables=[
                WorkloadTables(workload_type=workload_type, scale_factor=scale_factor)
                for workload_type, driver in drivers.items()
                for scale_factor in driver.scale_factors
            ]
        )
```

hyrisecockpit/api/app/status/model.py
```diff
@@ -19,14 +19,36 @@ def __init__(
...
-        """Initialize a benchmark status model."""
-        self.id: str = id
-        self.loaded_benchmarks: List[str] = loaded_benchmarks
+    def __init__(
+        self,
+        workload_type: str,
+        scale_factor: float,
+        loaded_tables: List[str],
+        missing_tables: List[str],
+        completely_loaded: bool,
```
heseding, line 31: I believe this is redundant as `completely_loaded == not missing_tables`.

hyrisecockpit/api/app/status/model.py
```diff
@@ -19,14 +19,36 @@ def __init__(
...
+        database_representation: Dict,
```
heseding:
```suggestion
        database_representation: Dict[str, str],
```

hyrisecockpit/api/app/status/model.py
```diff
@@ -19,14 +19,36 @@ def __init__(
...
+    ):
+        """Initialize a tables status model."""
+        self.workload_type: str = workload_type
+        self.scale_factor: float = scale_factor
         self.loaded_tables: List[str] = loaded_tables
+        self.missing_tables: List[str] = missing_tables
+        self.completely_loaded: bool = completely_loaded
+        self.database_representation: Dict = database_representation
```
heseding:
```suggestion
        self.database_representation: Dict[str, str] = database_representation
```

hyrisecockpit/api/app/status/service.py
```diff
@@ -35,15 +41,20 @@ def get_database_status(cls) -> List[DatabaseStatus]:
...
-        ]
+        workload_tables: List[WorkloadTablesStatus] = []
+        for database in response["body"]["workoad_tables"]:
+            workload_tables_status = [
+                TablesStatus(**status) for status in database["workoad_tables_status"]
+            ]
+            workload_tables.append(
+                WorkloadTablesStatus(database["id"], workload_tables_status)
+            )
+        return workload_tables
```
heseding: See [PEP 202](https://www.python.org/dev/peps/pep-0202/):
> List comprehensions provide a more concise way to create lists in situations where [...] nested loops would currently be used.
```suggestion
        return [
            WorkloadTablesStatus(
                database["id"],
                [
                    TablesStatus(**status)
                    for status in database["workoad_tables_status"]
                ],
            )
            for database in response["body"]["workoad_tables"]
        ]
```

hyrisecockpit/api/app/workload/interface.py
```diff
@@ -1,15 +1,25 @@
...
+    """Interface of a Workload."""
+
+    running: bool
+
+
+class DetailedWorkloadInterface(BaseWorkloadInterface):
     """Detailed interface of a Workload."""
 
-    weights: Dict[str, float]
+    running: bool
```
heseding: This should inherit from `WorkloadInterface` instead of `BaseWorkloadInterface` to remove the duplicate `running: bool`.

hyrisecockpit/api/app/workload/interface.py
```diff
@@ -1,15 +1,25 @@
...
+    supported_scale_factors: List
+    default_weights: Dict
```
heseding:
```suggestion
    supported_scale_factors: List[float]
    default_weights: Dict[str, float]
```

hyrisecockpit/api/app/workload/model.py
```diff
@@ -1,20 +1,56 @@
...
+
+    def __init__(
+        self,
+        workload_type: str,
+        frequency: int,
+        scale_factor: float,
+        weights: Dict[str, float],
+        running: bool,
+        supported_scale_factors: List,
+        default_weights: Dict,
```
heseding:
```suggestion
        supported_scale_factors: List[float],
        default_weights: Dict[str, float],
```

hyrisecockpit/api/app/workload/model.py
```diff
@@ -1,20 +1,56 @@
...
+    ):
+        """Initialize a detailed Workload model."""
+        self.running: bool = running
+        self.supported_scale_factors: List = supported_scale_factors
+        self.default_weights: Dict = default_weights
```
heseding:
```suggestion
        self.supported_scale_factors: List[float] = supported_scale_factors
        self.default_weights: Dict[str, float] = default_weights
```

hyrisecockpit/api/app/workload/model.py
```diff
@@ -1,20 +1,56 @@
...
+
+        super().__init__(workload_type, frequency, scale_factor, weights)
```
heseding: Should inherit from `Workload`.
```suggestion
class DetailedWorkload(Workload):
    """Model of a detailed Workload."""

    def __init__(
        self,
        workload_type: str,
        frequency: int,
        scale_factor: float,
        weights: Dict[str, float],
        running: bool,
        supported_scale_factors: List,
        default_weights: Dict,
    ):
        """Initialize a detailed Workload model."""
        self.supported_scale_factors: List = supported_scale_factors
        self.default_weights: Dict = default_weights

        super().__init__(workload_type, frequency, scale_factor, weights, running)
```

hyrisecockpit/api/app/workload/schema.py
```diff
@@ -1,25 +1,38 @@
...
     """Detailed schema of a Workload."""
 
-    weights = Dict(
+    running = Boolean(description="Running status of workload.",)
+
+
+class DetailedWorkloadSchema(BaseWorkloadSchema):
+    """Detailed schema of a detailed Workload."""
+
+    running = Boolean(description="Running status of workload.",)
```
heseding, line 30:
```suggestion
class DetailedWorkloadSchema(WorkloadSchema):
    """Detailed schema of a detailed Workload."""

```

hyrisecockpit/database_manager/database.py
```diff
@@ -59,12 +60,14 @@ def __init__(
 
         self._database_blocked: Value = Value("b", False)
         self._hyrise_active: Value = Value("b", True)
+        self._workload_drivers: Dict = Connector.get_workload_drivers()
```
heseding, line 63:
```suggestion
        self._workload_drivers: Dict[str, Any] = Connector.get_workload_drivers()
```
There could be a stronger type hint here (`Dict[str, Driver]`) if all workload drivers would inherit from a base `Driver`.

hyrisecockpit/database_manager/database.py
```diff
@@ -109,21 +112,39 @@ def get_queue_length(self) -> int:
...
-    def load_data(self, folder_name: str) -> bool:
+    def load_data(self, workload: Dict) -> bool:
         """Load pre-generated tables."""
-        return (
-            False
-            if self._worker_pool.get_status() != "closed"
-            else self._background_scheduler.load_tables(folder_name)
-        )
+        workload_type = workload["workload_type"]
+        scale_factor = workload["scale_factor"]
```
heseding, line 155: Why not use `workload_type` and `scale_factor` as parameters as done everywhere else instead of a dictionary?

hyrisecockpit/database_manager/database.py
```diff
@@ -109,21 +112,39 @@ def get_queue_length(self) -> int:
...
+        if workload_type not in list(self._workload_drivers.keys()):
```
heseding:
```suggestion
        if workload_type not in self._workload_drivers:
```
https://stackoverflow.com/a/1602944

hyrisecockpit/database_manager/database.py
```diff
@@ -109,21 +112,39 @@ def get_queue_length(self) -> int:
...
-    def delete_data(self, folder_name: str) -> bool:
+    def delete_data(self, workload: Dict) -> bool:
         """Delete tables."""
-        return (
-            False
-            if self._worker_pool.get_status() != "closed"
-            else self._background_scheduler.delete_tables(folder_name)
-        )
+        workload_type = workload["workload_type"]
+        scale_factor = workload["scale_factor"]
```
heseding, line 172: Why not use `workload_type` and `scale_factor` as parameters as done everywhere else instead of a dictionary?

hyrisecockpit/database_manager/database.py
```diff
@@ -109,21 +112,39 @@ def get_queue_length(self) -> int:
...
+        if workload_type not in list(self._workload_drivers.keys()):
```
heseding:
```suggestion
        if workload_type not in self._workload_drivers:
```
https://stackoverflow.com/a/1602944

hyrisecockpit/database_manager/database.py
```diff
@@ -148,8 +169,9 @@ def get_hyrise_active(self) -> bool:
         """Return status of hyrise."""
         return bool(self._hyrise_active.value)
 
-    def get_loaded_tables(self) -> List[Dict[str, str]]:
+    def get_loaded_tables_in_database(self) -> List[Dict[str, str]]:
         """Return already loaded tables."""
+        # TODO Adjust
```
heseding: Adjust what?

hyrisecockpit/database_manager/database.py
```diff
@@ -159,34 +181,37 @@ def get_loaded_tables(self) -> List[Dict[str, str]]:
         else:
             return [row[0] for row in rows] if rows else []
 
-    def get_loaded_benchmarks(self, loaded_tables) -> List[str]:
+    def _workoad_tables_status(self, tables_in_database) -> List:
```
heseding:
```suggestion
    def _workload_tables_status(self, tables_in_database) -> List:
```

hyrisecockpit/database_manager/database.py
```diff
@@ -159,34 +181,37 @@ def get_loaded_tables(self) -> List[Dict[str, str]]:
...
+                    "scale_factor": scale_factor,
+                    "loaded_tables": loaded_tables,
+                    "missing_tables": missing_tables,
+                    "completely_loaded": loaded,
+                    "database_representation": workload_tables,
+                }
+                workoad_tables_status.append(table)
+        return workoad_tables_status
+
+    def get_workoad_tables_status(self) -> List:
```
heseding:
```suggestion
    def get_workload_tables_status(self) -> List:
```

hyrisecockpit/database_manager/database.py
```diff
@@ -159,34 +181,37 @@ def get_loaded_tables(self) -> List[Dict[str, str]]:
...
         """Get loaded benchmark data."""
-        loaded_tables: List = self.get_loaded_tables()
-        loaded_benchmarks: List = self.get_loaded_benchmarks(loaded_tables)
-        return (
-            loaded_tables,
-            loaded_benchmarks,
-        )
+        tables_in_database: List = self.get_loaded_tables_in_database()
+        workoad_tables_status = self._workoad_tables_status(tables_in_database)
+        return workoad_tables_status
```
heseding:
```suggestion
        return self._workoad_tables_status(tables_in_database)
```

hyrisecockpit/database_manager/job/get_loaded_tables.py
```diff
@@ -15,7 +13,7 @@ def get_loaded_tables_for_scale_factor(
             results = cur.fetchall()
             all_loaded_tables = [row[0] for row in results] if results else []
             for table in tables:
-                if f"{table}_{benchmark}_{scale_factor}" in all_loaded_tables:
+                if table in all_loaded_tables:
                     loaded_tables.append(table)
```
heseding: A list comprehension is more concise.

hyrisecockpit/database_manager/manager.py
```diff
@@ -268,19 +266,15 @@ def _call_database_status(self, body: Body) -> Response:
...
         for database_id, database in self._databases.items():
-            loaded_tables, loaded_benchmarks = database.get_loaded_benchmark_data()
+            workoad_tables_status = database.get_workoad_tables_status()
             status.append(
-                {
-                    "id": database_id,
-                    "loaded_benchmarks": loaded_benchmarks,
-                    "loaded_tables": loaded_tables,
-                }
+                {"id": database_id, "workoad_tables_status": workoad_tables_status}
```
heseding: A list comprehension is more concise.

hyrisecockpit/database_manager/worker/task_worker.py
```diff
@@ -0,0 +1,84 @@
...
+                            worker_id,
+                            commited,
+                        )
+                    )
+                except Empty:
+                    continue
+                except (ValueError, ProgrammingError) as e:
+                    failed_queries.append((time_ns(), worker_id, str(task), str(e)))
+                except (DatabaseError, InterfaceError):
+                    task_queue.put(task)
```
heseding, line 80:
```suggestion
                except Empty:
                    continue
                except (ValueError, ProgrammingError) as e:
                    failed_queries.append((time_ns(), worker_id, str(task), str(e)))
                except (DatabaseError, InterfaceError):
                    task_queue.put(task)
                else:
                    succesful_queries.append(
                        (
                            endts,
                            latency,
                            benchmark,
                            scalefactor,
                            query_type,
                            worker_id,
                            commited,
                        )
                    )
```

hyrisecockpit/drivers/__default__/default_workload.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import List
+
+from hyrisecockpit.drivers.__default__.task_types import DefaultTask
+from hyrisecockpit.drivers.__default__.workload_reader import WorkloadReader
+
+
+class DefaultWorkload:
+    """Generates workloads from queries."""
+
+    def __init__(self, benchmark: str, scalefactor: float, query_path: str):
```
heseding, line 13:
```suggestion
    def __init__(self, benchmark: str, scale_factor: float, query_path: str):
```
I noticed its written `scale_factor` in the `app/` files and `scalefactor` here. We should consolidate the naming.

hyrisecockpit/drivers/__default__/driver.py
```diff
@@ -0,0 +1,133 @@
...
+        return workload.get(frequency, weights)
+
+    def _get_formatted_scalefactor(self, scalefactor: int) -> str:
+        """Get formatted scalefactor."""
+        if scalefactor == int(scalefactor):
+            formatted_scalefactor = str(int(scalefactor))
+        else:
+            formatted_scalefactor = str(scalefactor).replace(".", "_")
+
+        return formatted_scalefactor
```
heseding, line 51: The type hint for `scalefactor` does not make sense for the conditional expression.
```suggestion
    def _get_formatted_scalefactor(self, scalefactor: Union[int, float]) -> str:
        """Get formatted scalefactor."""
        if scalefactor == int(scalefactor):
            scalefactor = int(scalefactor)
        return str(scalefactor).replace(".", "_")
```

hyrisecockpit/drivers/__default__/driver.py
```diff
@@ -0,0 +1,133 @@
...
+        """Execute task of the query type."""
+        query = task["query"]
+        query = query.replace("[STREAM_ID]", str(worker_id))
+
+        not_formatted_parameters = task["args"]
+        formatted_parameters = self._get_formatted_parameters(not_formatted_parameters)
+        commited = True
+
+        endts, latency = self._execute_query(cursor, query, formatted_parameters)
+        return endts, latency, task["scalefactor"], task["query_type"], commited
```
heseding, line 133: What does the `commited` variable mean and why is it always `True`?
```suggestion

        endts, latency = self._execute_query(cursor, query, formatted_parameters)
        return endts, latency, task["scalefactor"], task["query_type"], True
```

hyrisecockpit/drivers/__default__/workload_reader.py
```diff
@@ -0,0 +1,29 @@
...
+    @classmethod
+    def get(cls, query_path: str) -> Optional[Dict[str, List[str]]]:
+        """Get a workload from the workload folder. Returns the query names and a list of all queries."""
+        queries = {}
+        for file_name in listdir(query_path):
+            query_type = file_name.split(".sql")[0]
+            query_file_path = f"{query_path}/{file_name}"
+            queries[query_type] = cls._read_query(query_file_path)
+
+        return queries
```
heseding:
```suggestion
        return {
            file_name.split(".sql")[0]: cls._read_query(f"{query_path}/{file_name}")
            for file_name in listdir(query_path)
        }
```

hyrisecockpit/drivers/connector.py
```diff
@@ -0,0 +1,58 @@
+"""Connector for cockpit and drivers."""
+from hyrisecockpit.drivers.job.job_driver import JobDriver
+from hyrisecockpit.drivers.tpcc.tpcc_driver import TpccDriver
+from hyrisecockpit.drivers.tpcds.tpcds_driver import TpcdsDriver
+from hyrisecockpit.drivers.tpch.tpch_driver import TpchDriver
+
+
+class Workload:
```
heseding, line 8: Type hints for the variables and return values of methods would increase readability and provide better static type checking.

hyrisecockpit/drivers/connector.py
```diff
@@ -0,0 +1,58 @@
...
+        return self.driver.get_default_weights()
+
+    def update(self, scale_factor, frequency, weights):
+        """Update workload object."""
+        if bool(weights):
+            self.weights = weights
+        self.scale_factor = scale_factor
+        self.frequency = frequency
+
+    def resert(self):
```
heseding:
```suggestion
    def reset(self):
```

hyrisecockpit/drivers/connector.py
```diff
@@ -0,0 +1,58 @@
...
+        self.frequency = 0
+        self.driver = driver
+
+    def get_default_weights(self):
+        """Return default weights."""
+        return self.driver.get_default_weights()
+
+    def update(self, scale_factor, frequency, weights):
+        """Update workload object."""
+        if bool(weights):
```
heseding:
```suggestion
        if weights:
```

hyrisecockpit/drivers/connector.py
```diff
@@ -0,0 +1,58 @@
...
+        self.frequency = frequency
+
+    def resert(self):
+        """Reset attributes."""
+        self.scale_factor = 0
+        self.weights = self.driver.get_default_weights()
+        self.frequency = 0
+
+
+class Connector:
```
heseding, line 37: Type hints for the variables and return values of methods would increase readability and provide better static type checking.

hyrisecockpit/drivers/connector.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @classmethod
+    def get_workload(cls):
+        """Return a dictionary with workload objects."""
+        return {
+            "tpch": Workload(TpchDriver()),
+            "tpcc": Workload(TpccDriver()),
+            "tpcds": Workload(TpcdsDriver()),
+            "job": Workload(JobDriver()),
+        }
```
heseding:
```suggestion
        return {key: Workload(value) for key, value in get_workload_drivers().items()}
```

hyrisecockpit/drivers/job/job_driver.py
```diff
@@ -0,0 +1,58 @@
...
+        self._benchmark_type: str = "job"
+        self._table_names: List[str] = tables
+        self.scale_factors = [1.0]
+        self._scale_factor_query_path = {1.0: "job_1"}
+        self._default_driver: DefaultDriver = DefaultDriver(
+            self._query_path,
+            self._scale_factor_query_path,
+            self._benchmark_type,
+            self._table_names,
+        )
```
heseding: We might be better off using a `Driver` class and creating instances for JOB, TPC-H, etc., and something like a `DynamicDriver` class for TPC-C instead of hardcoding configurations in all `__init__` methods.

hyrisecockpit/drivers/job/job_driver.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def get_delete_queries(self, scalefactor) -> Dict[str, Tuple]:
+        """Generate delete table queries."""
+        return self._default_driver.get_delete_queries(scalefactor)
+
+    def execute_task(
+        self, task, cursor, worker_id
+    ) -> Tuple[int, int, float, str, bool]:
+        """Generate execute task queries."""
+        return self._default_driver.execute_task(task, cursor, worker_id)
```
heseding, line 57: All of this code is shared between JOB, TPC-H, and TPC-DS. We should remove the duplicate code (inherit from a parent `Driver` if you want to keep the classes or create instances of JOB, TPC-H, and TPC-DS drivers from a `Driver` class).

hyrisecockpit/drivers/tpcc/tpcc_driver.py
```diff
@@ -0,0 +1,95 @@
...
+            "order_status": 0.04,
+            "payment": 0.43,
+            "new_order": 0.45,
+        }
+
+    def generate(self, scalefactor, frequency, weights) -> List[DefaultTask]:
+        """Generate tpch tasks."""
+        self._parameter_generator.apply_scalefactor(scalefactor)
+        tasks = self._parameter_generator.generate_transactions(frequency, weights)
+        return tasks
```
heseding:
```suggestion
        return self._parameter_generator.generate_transactions(frequency, weights)
```

hyrisecockpit/drivers/tpcc/tpcc_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        commited = True
+        startts = time_ns()
+        try:
+            self._transaction_handler.execute_transaction(
+                cursor, transaction_type, scalefactor, parameters
+            )
+        except SerializationFailure:
+            commited = False
+            cursor.reset()
+        endts = time_ns()
```
heseding: We shouldn't measure the cursor reset time.
```suggestion
            endts = time_ns()
            commited = False
            cursor.reset()
        else:
            endts = time_ns()
```

hyrisecockpit/server.py
```diff
@@ -42,16 +40,13 @@ def start(self) -> None:
...
-            validate(instance=request, schema=request_schema)
-            func, schema = self._calls[request["header"]["message"]]
-            if schema:
-                validate(request["body"], schema=schema)
-            return func(request["body"])
-        except ValidationError:
-            return get_response(400)
-        except KeyError:
+
+        # TODO remove validation schema and call not found. We handle all this stuff in flask.
```
heseding, line 44: A server should still validate that requests are at least readable and drop them if they are not. This makes the server more robust and is in my opinion a super low hanging fruit.

hyrisecockpit/workload_generator/generator.py
```diff
@@ -76,81 +72,80 @@ def _init_server(self) -> None:
...
-            response = get_response(404)
-        else:
-            response = get_response(200)
-            response["body"]["folder_name"] = folder_name
+        workload_type: str = body["workload_type"]
+        workload = self._workloads.get(workload_type)
+        if workload is None:
+            return get_response(404)
+        workload.running = False
+        workload.resert()
```
heseding:
```suggestion
        workload.reset()
```

hyrisecockpit/api/app/status/service.py
Alexander-Dubrawski: I think in this case the classic for loop is just easier to read.

hyrisecockpit/drivers/__default__/driver.py
Alexander-Dubrawski: Please have a look at the description of this PR under *TPCC implemeentation*. Since tpcds, tpch and job are just reading transaction there can't be a conflict and so the commit is always successful.

hyrisecockpit/server.py
Alexander-Dubrawski: Because this server only communicates with the flask application and this application is well testet I don't see the need to validate the message. For development porpoise I think it is even better to let the application crash if a message is not valid to find the bug quickly. For me a seperate validation makes sense if we don't know who is communicating with it.

hyrisecockpit/drivers/job/job_driver.py
```diff
@@ -0,0 +1,57 @@
...
+from hyrisecockpit.drivers.job.job_meta_information import default_weights, tables
+
+
+class JobDriver:
+    """job driver."""
+
+    def __init__(self):
+        """Initialize a job driver."""
+        # TODO Move queries to driver folder
+        self._query_path: str = f"{Path(__file__).parent.parent.parent}/workload_generator/workloads"
```
Alexander-Dubrawski, line 17: Why are we using 'Path(__file__)' here? What is the benefit in comparison to abspath(getcwd())?

PeterTsayun: `getcwd()` returns a directory of the `__main__` file, whereas `Path` returns the path of the current file. So, `Path` always returns the same result, whereas the output of the `getcwd()`-function depends on the location of the file, where the application was started.

hyrisecockpit/drivers/job/job_driver.py
Alexander-Dubrawski: I think it is cleaner and easier to understand if we have a specific driver for every single workload. Because workload/benchmarks can be very specific. Some benchmarks that could be implemented in the future can be dynamic but still very different to the tpcc benchmark. Moreover it is easier to have different variants of one benchmark if we have a single driver for every benchmark. So we are mor flexible for future implementations if we have a clear separation of logic on the benchmark level.

hyrisecockpit/drivers/job/job_driver.py
Alexander-Dubrawski: If we really wan't to group the Workloads together I think we shouldn't call it Driver. Maybe DefaultOLAPDriver. The Tpcc driver should be untouched in my opinion.

hyrisecockpit/drivers/job/job_driver.py
heseding: > I think it is cleaner and easier to understand if we have a specific driver for every single workload.

I disagree, it would make more sense for me to have a class for each type of benchmark, i.e. static or dynamic (or whatever there will be in the future), and create instances representing the concrete benchmarks.
This would require only one class (`StaticDriver`) for TPC-H, TPC-DS, and JOB and only one class (`DynamicDriver`) for TPC-C.
As this results in fewer LOC, I am sure it would be more readable.

> Because workload/benchmarks can be very specific.

I agree.

> Some benchmarks that could be implemented in the future can be dynamic but still very different to the tpcc benchmark.

https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

> Moreover it is easier to have different variants of one benchmark if we have a single driver for every benchmark.

How would that not be possible by instantiating a driver again?
(Is that even a use case?)

> So we are mor flexible for future implementations if we have a clear separation of logic on the benchmark level.

What do you mean by clear separation of logic?
I don't think repeating code in multiple classes improves flexibility for future implementations.

hyrisecockpit/drivers/job/job_driver.py
heseding: Sounds good.

hyrisecockpit/server.py
heseding: > Because this server only communicates with the flask application and this application is well testet I don't see the need to validate the message.

The communication between flask and the dbm/wlg is not well tested.

> For development porpoise I think it is even better to let the application crash if a message is not valid to find the bug quickly.

Sure, we could add a `validate: bool` flag that could be easily be switched off when developing.

Besides, I don't see how this change is crucial to adding TPC-C?

PR #673, opened 2020-07-07
B/load already loaded plugin

### LINKED ISSUE #591: Hyrise crashes when trying to load already existing plugin
**Please describe the bug:**  
If we try to load the same plugin twice the hyrise crashes wit the exception: 

IMAGE 2020-05-06 11:11:17

1. Add a plugin 
2. Add the same plugin again

**Affected Component(s):**  
`DatabaseManager`, `Frontend`, ...

**Additional context:**  
We should check whenever we activate a plugin if it is already activated. We can do this in the frontend or backend.

PeterTsayun:
# Resolves #591

The Hyrise instance crashes when loading already loaded plugin. For that reason, now we are checking if the plugin to be activated has already been loaded.

No feature added.

**Affected Component(s):**  
`Database`

**Additional context:**


Bensk1:
How can we load an already loaded plugin? Does the frontend allow that?

Alexander-Dubrawski (approved):

PeterTsayun:
> How can we load an already loaded plugin? Does the frontend allow that?

After activation of the plugin, the frontend asks the backend which plugins are activated. So, when the activation of the plugin takes too much time,  it could happen, that the new plugin is shown as not activated. We should discuss with the frontend team, how we would handle this case.

PR #674, opened 2020-07-08
B/add different job managers

### LINKED ISSUE #664: Missing table when loading the benchmark
**Please describe the bug:**  
Sometimes, one table (`lineitem`) is missing and the cockpit remains in the blocked state.

1. Choose TPCH 1.0 as default tables in `.env`
2. Start Cockpit
3. Add database

The cockpit should load all tables for the database.

**Affected Component(s):**  
`BackgroundScheduler`

**Additional context:**  
It seems to be a problem in the hyrise. For more information see:
https://github.com/hyrise/hyrise/issues/2180

Alexander-Dubrawski:
# Resolves #664 

Like in https://github.com/hyrise/hyrise/issues/2180 described we can not easily load and delete tables while asking at the same time for the meta information. In my opinion, a simple solution for the moment is to stop requesting the meta-information while loading/ deleting the tables. For that reason, I split the jobs in the database object into three categories: 

* background jobs (that are executed every n second. For example getting the system data)
* asynchronous jobs (that are executed once but the client don't need an answer. For example load/delete tables)
* synchronous jobs (that are executed once and the client needs an answer. For example getting the plugins)

All the background jobs are executed with the Background scheduler (https://apscheduler.readthedocs.io/en/stable/index.html). All the asynchronous jobs are executed with simple python threads (https://docs.python.org/3/library/threading.html#module-threading). Every time we are loading/deleting the tables we are stopping the Background scheduler (We are getting al the meta-information with the background jobs). If the tables are loaded/deleted we are starting the Background scheduler again. 

To implement the solution and keep the coherence small I wrote three job handlers:

* background job handler
* asynchronous job handler
* synchronous job handler

All the handlers are defined in their own classes. The database object has an object from every manager and delegates the requests to the right one. 

image

**Affected Component(s):**  
`DatabaseManager`, `Database`, `jobs`

**Additional context:**  
Because this is a bigger change in the architecture of the database object I decided that it would be beneficial to discuss it in a draft first.


hyrisecockpit/database_manager/job_manager/background.py
```diff
@@ -0,0 +1,133 @@
...
+            ),
+        )
+
+    def start(self) -> None:
+        """Start background scheduler."""
+        self._scheduler.start()
+
+    def close(self) -> None:
+        """Close background scheduler."""
+        self._scheduler.shutdown()
```
PeterTsayun, line 128: Shouldn't we stop background jobs as well before stopping the scheduler?

Alexander-Dubrawski: By default the scheduler should shutdown all its jobs (and waits for there execution to be finished). 

https://apscheduler.readthedocs.io/en/stable/userguide.html#shutting-down-the-scheduler

Alexander-Dubrawski:
Implemented in #752 #757 and #756 .

PR #682, opened 2020-07-14
f/681 improve linechart plugin logs

### LINKED ISSUE #681: improve plugin log messages in linecharts
It is hard to read the plugin log messages at the moment, because they are moving and really small. 
The plugin messages should be activated by clicking on the event line. The message should stay in on place and show the whole text via a scrollable and half transparent box. 

**Affected Component(s):**  
frontend, linecharts

MaxSchneider1337:
# Resolves #681

improve the linechart plugin logs and shows them bigger and static.

**Affected Component(s):**  
Frontend

**Additional context:**  
@Bensk1


hyrisecockpit/frontend/src/components/container/MetricTile.vue
```diff
@@ -22,6 +22,13 @@
         </v-row>
       </v-container>
     </v-card-title>
+    <plugin-log-popup
+      v-if="showPluginPopup"
+      :current-plugin-log="currentPluginLog"
+      :current-plugin-log-database-color="popupColor"
+      :current-plugin-log-metric-id="tileDatabase + metric"
```
schTi: those prop names are a bit too long in my opinion. You could remove the `current-plugin-log` prefix as I see it

hyrisecockpit/frontend/src/components/container/MetricsComparisonTable.vue
```diff
@@ -49,7 +51,9 @@ import {
   onMounted,
   computed,
   provide,
+  reactive,
```
schTi: I dont see a reason why to import this

hyrisecockpit/frontend/src/components/plugins/PluginLogPopup.vue
```diff
@@ -0,0 +1,110 @@
...
+      data-id="chart-plugin-log-alert"
+      elevation="20"
+      :color="currentPluginLogDatabaseColor"
+    >
+      <div v-if="!!currentPluginLog">
+        <div class="header-row">
+          <div class="log-header">
+            {{ currentPluginLog.header }}
+          </div>
+          <v-btn @click="onClose()" icon>
```
schTi: I know you like this callback-stuff. But can you please use `$emit('close')` here instead and just call the onClose function `@close` on the parent? Then you dont have to pass the whole function down to the child

schTi (commented):
Can you please add some screenshots here? Or tell me what the benefit of this PR is? I am not to sure about it

MaxSchneider1337:
> Can you please add some screenshots here? Or tell me what the benefit of this PR is? I am not to sure about it

talked to @Bensk1, he wanted to have a better presentation of the plugin log text, when the user clicks in the linecharts. I will add Screenshots asap.

schTi (approved):
:fire:

monasobh:
I like that the information is in a see-through box but maybe you could add a default position, so that it is on the border top or something, so the user doesn't have to drag it up?!  
Screenshot 2020-07-29 at 10 09 00

MaxSchneider1337:
> I like that the information is in a see-through box but maybe you could add a default, so that its on the border right or something?
> Screenshot 2020-07-29 at 10 09 00

do you mean a default position for the box?

monasobh:
Also, could you please resize the font. I think it would be good to have the same size for the log messages as the "Aggregation: 1 second" (class="caption primary--text").

MaxSchneider1337:
> I like that the information is in a see-through box but maybe you could add a default, so that it is on the border top or something, so the user doesn't have to drag it up?!
> Screenshot 2020-07-29 at 10 09 00

tried to find a good position, but 

> I like that the information is in a see-through box but maybe you could add a default, so that it is on the border top or something, so the user doesn't have to drag it up?!
> Screenshot 2020-07-29 at 10 09 00

It's really hard to find a good position for this, because putting it on the top of the tile looks really on, when comparing 2 or more databases. I would suggest putting it in the middle, because the graph is not as disturbing as text in the background.
image

monasobh:
> It's really hard to find a good position for this, because putting it on the top of the tile looks really on, when comparing 2 or more databases. I would suggest putting it in the middle, because the graph is not as disturbing as text in the background.
> image

Then maybe the see-through is not the best solution. I think both backgrounds, the text in the background and the graph-lines in the background are a bit disturbing. Can you make a screenshot with white background and not see-through?

MaxSchneider1337:
> > It's really hard to find a good position for this, because putting it on the top of the tile looks really on, when comparing 2 or more databases. I would suggest putting it in the middle, because the graph is not as disturbing as text in the background.
> > image
> 
> Then maybe the see-through is not the best solution. I think both backgrounds, the text in the background and the graph-lines in the background are a bit disturbing. Can you make a screenshot with white background and not see-through?

I don't think removing the see through effect is a good solution, because a normal box hides a lot of information and makes it hard to see the graph

monasobh:
> I don't think removing the see through effect is a good solution, because a normal box hides a lot of information and makes it hard to see the graph

I'm just thinking of alternatives because the log text on dark db colors is really hard to read on graph lines.

Bensk1:
If it's not too much effort, we could try it without transparency or at least with (much) less transparency?

monasobh:
My suggestion would be something like that
Screenshot 2020-07-29 at 19 40 54

Bensk1:
That looks better. Could we move the close button a bit further to the right? Also, how about changing the font size for the header (CompressionPlugin [...]:) to the font size of the Y-axis which should be a bit bigger?

monasobh:
@MaxSchneider1337 will you implement the changes?

MaxSchneider1337:
> @MaxSchneider1337 will you implement the changes?

@monasobh @Bensk1 I really don't like this version, because on low resolutions or when two ore more dbs are connected the log message hides more than half of the graph. Thats why i made them transparent in the first palce.

monasobh:
> @monasobh @Bensk1 I really don't like this version, because on low resolutions or when two ore more dbs are connected 
the log message hides more than half of the graph. Thats why i made them transparent in the first palce.

I think its much more useful to shortly read the log message on white background and then close the window to see the full graph again instead of not being able to read the log message at all while not being able to see the graph clearly. 

pluginlog

monasobh:
pluginlog2

monasobh:
Also, imagine having "Aggregation: 1 Minute", then there would be a lot more dots and make the message really hard to read.

Bensk1:
Is the box only closed when the `x` is clicked or also when a click elsewhere is registered?

monasobh:
> Is the box only closed when the `x` is clicked or also when a click elsewhere is registered?

Currently only when the x is clicked but it's also possible to close it when clicking anywhere. 

@Bensk1 I just talked to Max and we agreed on the white background.

Bensk1:
> Currently only when the x is clicked but it's also possible to close it when clicking anywhere.

I think that would be good.

> @Bensk1 I just talked to Max and we agreed on the white background.

Great. I think we all agree that the way this is displayed could be improved but given the available time frame we should go for an acceptable solution.

monasobh:
Like this? Screenshot 2020-07-31 at 16 01 39

Bensk1:
Works for me.

monasobh:
> > Currently only when the x is clicked but it's also possible to close it when clicking anywhere.
> 
> I think that would be good.

Sorry, I was mistaken. This is not possible with Vuetify on this component.

PR #695, opened 2020-07-15
F/fix-workload-settings-and-equalizer

kathwill:
# Resolves #number

After updating the dependencies the frequencies and the equalizer were not shown properly. Moreover, after reloading the page the started workloads were not marked in the checkboxes.

**Affected Component(s):**  
`Frontend`

**Additional context:**  
This is how the workload settings look before the fix:
FrequenciesBug
EqualizerBug


schTi (approved):
Thanks for the fix!

PR #704, opened 2020-07-18
f/676-fix-db-value-bug

### LINKED ISSUE #676: Memory capacity and number of cpus in top bar is 0
**Please describe the bug:**  
After adding a database the system data values for Memory Capacity and CPUs of the added database is 0 (in the top bar). After reloading the page the values are updated.

1. Add a database
2. Go to the Overview or Comparison page
3. Look at Memory Capacity and CPUs for the added database in the top bar

After receiving the values for Memory Capacity and CPUs the values in the top bar should reload.

**Screenshot(s):**  
Screenshot 2020-07-10 at 16 10 01

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #676

There was a bug, that the database details for cpu data did not update after adding a new database. 
Unfortunately it takes some time, until the cpu data is loaded in the backend. So it will still show 0 on initial render. Nevertheless there was a bug, that the values were never updated. When the backend finally returns the correct cpu data it will now immediately be used in the frontend.

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/src/controller/databaseController.ts
+++ b/hyrisecockpit/frontend/src/controller/databaseController.ts
@@ -28,8 +28,8 @@ export function useDatabaseController(): DatabaseController {
     }
   });
 
-  eventBus.$on("CPU_DATA_CHANGED", (data: any) => {
-    if (allDatabasesExist(Object.keys(data))) {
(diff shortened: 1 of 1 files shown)


MaxSchneider1337 (approved):
N 🍦

PR #705, opened 2020-07-18
f/687-scroll-plugin-logs

### LINKED ISSUE #687: Change sorting in plugin log
At the moment the log messages in the plugin log are sorted by time in ascending order. Because of that the user needs to scroll to the end of the log to see the current log messages.
PluginLog

Sorting the plugin log messages in descending order solves the problem.

**Affected Component(s):**  
`Frontend`

schTi:
# Resolves #687

This PR adds auto scrolling to the latest plugin log entry

**Affected Component(s):**  
`Frontend`

### DIFF
--- a/hyrisecockpit/frontend/src/components/plugins/PluginsLog.vue
+++ b/hyrisecockpit/frontend/src/components/plugins/PluginsLog.vue
@@ -1,7 +1,7 @@
 <template>
   <v-expansion-panels flat>
     <v-expansion-panel>
-      <v-expansion-panel-header flat>
+      <v-expansion-panel-header flat @click="scrollToBottom(300)">
         <div class="log" data-id="plugin-log">Plugin log messages</div>
       </v-expansion-panel-header>
       <v-expansion-panel-content class="content">
@@ -22,10 +22,10 @@
 import {
   defineComponent,
   SetupContext,
-  onMounted,
   computed,
   Ref,
   ref,
+  watch,
 } from "@vue/composition-api";
 
 interface Props {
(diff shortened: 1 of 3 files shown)


MaxSchneider1337 (commented):
how is the performance??? this has to be sorted every second, which seems to be very heavy

I think the array simply has to be reversed

MaxSchneider1337:
> how is the performance??? this has to be sorted every second, which seems to be very heavy
> 
> I think the array simply has to be reversed

maybe use `array.reverse()` or some functionality of the vuetify textbox

schTi:
Can we ensure that those logs are sorted oldest > latest when returned from backend?

MaxSchneider1337:
> Can we ensure that those logs are sorted oldest > latest when returned from backend?

i think so, because the current textbox always shows them oldest to latest. I also think, that this direction (oldest to latest) ist correct. The textbox should auto-scroll to the bottom all the time.

MaxSchneider1337:
> > Can we ensure that those logs are sorted oldest > latest when returned from backend?
> 
> i think so, because the current textbox always shows them oldest to latest. I also think, that this direction (oldest to latest) ist correct. The textbox should auto-scroll to the bottom all the time.

at least that would be my suggestion. Could be much harder to implement

schTi:
> > Can we ensure that those logs are sorted oldest > latest when returned from backend?
> 
> i think so, because the current textbox always shows them oldest to latest. I also think, that this direction (oldest to latest) ist correct. The textbox should auto-scroll to the bottom all the time.

Yeah. Maybe you are right. I will have a look how this would work

schTi:
Added auto scroll to the bottom log entry. If you scroll manually to the top, it will stop.

MaxSchneider1337 (approved):
Sorry for the additional work, but it looks really nice now

PR #721, opened 2020-07-28
f/add-segment-heatmap

schTi:
# Resolves #number

This PR adds a new heatmap showing segment configuration data.

**Affected Component(s):**  
`Frontend`


schTi:
Unfortunately I was a bit blocked by the related backend PR. Nevertheless I finally added the segment configurations graph. It shows for every segment of a table the related encoding type or order mode. To switch between those metrics, you can use the button in the top right corner. The `bar-code icon` indicates encoding type and the `sorting icon` indicates order mode. All encodings/ order modes will be mapped to a unique color in the heat map. 

_NOTE: Those encodings are mocked for testing purposes._

Bildschirmfoto 2020-08-08 um 13 56 13
Bildschirmfoto 2020-08-08 um 13 56 33

How does this look to you @Bensk1 @Bouncner ? I cannot address too many issues on here any more. But if you want to have different icons / labels / colors, just hit me up. Otherwise I am going to merge this on Monday.

MaxSchneider1337 (approved):
nice thanks for the work

Bouncner:
Can we have categorical values with a better color scheme? Continuous does really make sense here.

Bouncner:
Is the sorting shown within the heat map? It’s not really necessary to have a second map for that.

No worries if the answer is “doable, but takes 10h of time; won’t do”.

schTi:
Thanks for the comments. Showing the order mode directly in the graph, would be possible for sure, but it would require parsing the data twice with the current implementation. Additionally the order mode data will be empty for the most time, so I will refrain from doing this now. Should not be a big deal switching between the data, if you want to see the order mode.

schTi:
Related to the color scheme, I will have a look, if I can make it discrete. In general they already are, but plotly applies a continuous color bar per default.

Bensk1:
I agree, a discrete color scheme would be nice.

schTi:
Bildschirmfoto 2020-08-10 um 16 11 44

Worked! This is going to be the final state, here

Bouncner:
> Showing the order mode directly in the graph, would be possible for sure, but it would require parsing the data twice with the current implementation.

Can you elaborate a bit what needs to be done exactly? It's probably up to somebody else (possible someone who was not part of this project)to implement it. So any help is appreciated here.

schTi:
The data provided by backend is splitted into encoding_type and order_mode on top level. Depending on which key is currently selected via button I am only parsing the necessary sub tree. To get all of the data at the same time I would need to parse both subtrees every time. No rocket science, but some additional work

Bouncner:
And putting text in the heat map should be straight forward?

schTi:
Yes for sure. You just need to adapt the hover template, which was already done multiple times in our application

PR #724, opened 2020-07-29
F/586-disable-adding-a-database-when-workers-are-running

### LINKED ISSUE #586: Disable adding a database when a workload is running
Currently it is possible to add a database, start a workload and then add another database. So there will be one instance with a running workload and one that might not even have the right data loaded. It is then not possible to load or delete tables because a workload is running.

Adding a database should be disabled when a workload is running. A warning dialog should be shown that might even has the option to stop the workload and then add a database.

kathwill:
# Resolves #586 

When you try to add a database while the workers are running, this will now be impossible and a warning will be shown with the opportunity to stop the running workload. So you will not have to leave the dialog, but you can still decide to cancel adding a database.

**Affected Component(s):**  
`Frontend`

**Additional context:**  
AddDatabaseStopWorker2


hyrisecockpit/frontend/src/components/databases/AddDatabase.vue
```diff
@@ -35,6 +35,34 @@
...
+          dense
+        >
+          <v-row>
+            <v-col class="grow pr-0 py-0"
+              >Stop workers before adding databases.</v-col
+            >
+            <v-col class="shrink mr-2 pl-0 py-0">
+              <v-tooltip right>
+                <template v-slot:activator="{ on }">
+                  <v-btn
```
schTi, line 52: Can you make the button white? Looks a bit lost in the alert with the current design

schTi (changes requested):

schTi (approved):

PR #725, opened 2020-07-29
Add a simple start script

PeterTsayun:
# Resolves #number

This PR adds a simple Python script for starting the cockpit components (backend, manager, generator). Every component is started in its own process. The stdout and stderr are captured via a pipe and logged (For more information see https://docs.python.org/3/howto/logging.html). The stdout is logged on the INFO level the stderr on the WARNING level. If the script is closed by a keyboard interrupt, we send a SIGINT to the components. If the components have an error and do not react we make sure to clear the ports via *fuser* (https://linux.die.net/man/1/fuser).

Add a simple script to start the cockpit in its developer environment.

Screenshot 2020-08-12 at 16 01 47

**Additional context:**

*Docker:*  
In the future, we can further simplify the docker setup. Instead of running the backend, manager, and generator all in a different container, we can now run them in one. That's makes it easier to manage the .env because we do not need to use the container URL (We can simply use the local URL). So the development setup is more straight forward because we do not need to adjust the .env (remove the container).

*Entry point:* 
If you want to use the new *cockpit* entry point to start the script in pipenv please run: 
1. pipenv --rm
2. pipenv --three install
3. pipenv update --clear


system_tests/setup.py
```diff
@@ -0,0 +1,25 @@
+"""Module for starting the cockpit components."""
+
+from threading import Event
+
+from system_tests.cockpit_backend import CockpitBackend
```
Alexander-Dubrawski: I'm not sure if we should use the components form the system tests because we are not using the default .env setup there.

system_tests/setup.py
```diff
@@ -0,0 +1,25 @@
...
+try:
+    print("Starting components...")
+    backend.start()
+    manager.start()  # type: ignore
+    generator.start()  # type: ignore
+    print("Cockpit is running...")
+    Event().wait()
+except KeyboardInterrupt:
+    print("Shutting down...")
+    manager.shutdown()  # type: ignore
```
Alexander-Dubrawski: In the system tests we are sending a SIGINT to the processes. In some cases if the program in the processes has a failure it can happen that the signal is not reaching it. Maybe we can double check with fuser (https://linux.die.net/man/1/fuser) that the process is really closed and more importantly the port are free.

system_tests/setup.py
```diff
@@ -0,0 +1,25 @@
+"""Module for starting the cockpit components."""
+
+from threading import Event
+
+from system_tests.cockpit_backend import CockpitBackend
+from system_tests.cockpit_generator import CockpitGenerator
```
Alexander-Dubrawski: If I remember correctly, we are writing the STDOUT and STDERR in a file in the system tests. Maybe we should just display it directly in the terminal here. For that it would be easier not to use the components from the system tests but instead write just a few lines of code here in the file to start the components.

Alexander-Dubrawski:
implemented in  #747 .

PR #729, opened 2020-07-30
F/726-rename-table-names-in-frontend

### LINKED ISSUE #726: Rename table names in frontend
The shown table names in the heatmap and treemap of the comparison page are difficult to read.

Mapping the table names to selected labels will solve the problem.

**Affected Component(s):**  
`Frontend`

kathwill:
# Resolves #726 

The shown table names in the heatmap and treemap of the comparison page are difficult to read.

By mapping the table names to selected labels the readability is improved.

**Affected Component(s):**  
`Frontend`

**Additional context:**  
RenameTableNames1
RenameTableNames2
RenameTableNames3


hyrisecockpit/frontend/src/meta/names.ts
```diff
@@ -0,0 +1,16 @@
+export function changeTableName(table: string): string {
+  const words = table.split("_");
```
Bouncner: Do you have an example input here? I am worried that this could be a little bit to unspecific and might catch cases it shouldn’t.

hyrisecockpit/frontend/src/meta/names.ts
```diff
@@ -0,0 +1,16 @@
...
+      words[1].toUpperCase() +
+      " SF " +
+      words[2] +
+      "." +
+      words[3] +
+      ")"
+    );
+  }
+  return words[0] + " (" + words[1].toUpperCase() + " SF " + words[2] + ")";
+}
```
schTi: Instead of using direct string concatenation, can you use template string? (see https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/template_strings)

schTi (changes requested):

hyrisecockpit/frontend/src/meta/names.ts
kathwill: lineitem_tpch_0_1

Bouncner: Hmm, the Join order benchmark has a table called `comp_cast_type.csv`, almost reaching the threshold. To be more future proof, we could add a safeguard that expects `tpch`/`tpcds`/`job` in the string.

@Bensk1 : do you know any benchmark tables that might already screw this up?

Bensk1: No, I don't know anything right now but I agree that this could easily be the case.

kathwill: I added a safeguard for these cases now

hyrisecockpit/frontend/src/meta/workloads.ts
```diff
@@ -35,3 +36,23 @@ export function getDisplayedFromTransferred(transferred: string): string {
     (workload) => workload.transferred === transferred
   )!.displayed;
 }
+
+export function getTableName(table: string): string {
+  if (isInTestMode) return table;
+
+  const index: number = Math.max(
+    table.indexOf("tpch"),
```
schTi, line 42: I don't think that this will hold for the described cases. You should just return the plain table name, if none of the already defined workload names is included in the table name string. Otherwise our application will crash when the table name does not contain a '_'. This is what happened before I added the test exception

schTi (changes requested):

hyrisecockpit/frontend/src/meta/workloads.ts
kathwill: But there is no table without a workload name included. Or do you have an example?

Bouncner: There might not be one right now, but the point of my request is that we should ensure that future workloads should still work even when they have multiple `_` them. Nobody knows how these table names will look like.

kathwill: If you mean `comp_cast_type.csv`, we receive this as `comp_cast_type_job_1` in the frontend. I can still add a return for your mentioned case, but I don't know if this is necessary.

Bouncner: No, I am talking about a file-based benchmark that we are going to add in some point in the future. Basically no one knows what the table names are going to be.

kathwill: @Bouncner Alright. I will change that. (My former answer was for @schTi, I didn't see your answer until now.)

schTi (approved):

PR #746, opened 2020-08-11
Docker/check_for_cached_tables

Bouncner:
When setting up a naked installation of the cockpit, the cached_tables folder is missing and data thus cannot be loaded (that the UI doesn't complain might become another issue).
One needs to check the hyrise container logs to see that the bin files are missing.

This is just a first try (I am a novice with Docker) to get the discussion going. 
I would like to get @cH3n7i 's feedback AFTER he finished TI2.
Another thing I am curious about is whether we should `wget` the files from "somewhere" (let's say archive.org) if the folder does not exist. I'd like to be able to install the cockpit everywhere without access to the VMs storing the bins or the HPI network at all.

Example is stolen from: https://stackoverflow.com/questions/48357244/docker-check-if-file-exists-in-healthcheck

**Affected Component(s):**  
`Docker`

### DIFF
--- a/docker-compose.deployment.yml
+++ b/docker-compose.deployment.yml
@@ -42,6 +42,8 @@ services:
 
   hyrise:
     image: udopigorsch/hyrise:bp1920
+    healthcheck:
+      test: bash -c [ -f ./cached_tables ]
     volumes:
       - ./cached_tables:/usr/local/hyrise/cached_tables
     ports:


Bouncner:
@Alexander-Dubrawski : I guess will merge #747 in the very near future. This PR can be closed, right?

Alexander-Dubrawski:
Implemented in  #747 ,

PR #747, opened 2020-08-18
Improve docker setup

Alexander-Dubrawski:
# Improve Docker setup

* This Pr implements a clear separation between the docker environment variables and the native ones. This is needed because in docker the containers need to know the names of the other containers to talk to each other. On the other hand, if we use the names of the containers per default the normal developer/native setup is not working.
* Before this pr an image was build and pushed to the Docker Hub in every commit. Since the build of the frontend image is depending on the host IP, automated updates of the images are with our architecture at the moment not doable. So I removed watchtower an the automatic push of the frontend and backend images to Docker Hub.   
* The Hyrise Image is building the binary tables. 
* All the backend components (backend, generator, manager) are now in one image. That reduces the complexity of the setup. The three components can be started via a python script.  
* Moreover I adjusted the Docker start guide https://github.com/hyrise/Cockpit/wiki/Docker-Setup


Bouncner:
Will take a deeper look at a later point in time. But that sounds all good to me (having a low docker knowledge though).

One thing we could do is loading the table data from a HPI share, if that does not succeed from a dropbox link, and if even that does not succeed load it from archive.org.

docker-compose.yml
```diff
@@ -40,15 +21,23 @@ services:
...
     image: hyrise:bp1920
     build: 
       context: .
       dockerfile: ./hyrise/Dockerfile
-    volumes:
-      - ./cached_tables:/usr/local/hyrise/cached_tables
     ports:
-      - "5432:5432"
\ No newline at end of file
+      - "6432:5432"
```
Bouncner, line 41: I wouldn't know if the first or the second port shall be used within the UI. Maybe use `"5433:5433"`?

hyrise/Dockerfile
```diff
@@ -1,24 +1,31 @@
 # https://github.com/hyrise/hyrise/wiki/Docker-Image
 FROM hyrise/opossum-ci:20.04
-LABEL maintainer="upigorsch@me.com"
 ENV OPOSSUM_HEADLESS_SETUP=true
 
 # From now on, we're doing additional stuff that would be done manually otherwise
 RUN apt-get update
+RUN apt-get -y install p7zip \
```
Bouncner: I'd guess even alpine has `zip` and `unzip` to use directly. As long as we don't need special features, I would keep the installed software small.
What might be interesting nonetheless (don't know if alpine has it out of the box) is better compression such as `xz`. We could compress it with the highest levels (done once, takes quite a while) and then have a much smaller file to transfer.

hyrisecockpit/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.8.5-slim
+
+WORKDIR /usr/local/Cockpit
+COPY . . 
+ADD ./hyrisecockpit/workload_generator/workloads /usr/local/lib/python3.8/site-packages/hyrisecockpit/workload_generator/workloads   
+RUN pip3.8 install -r requirements.txt \
+    && pip3.8 install .
```
Bouncner: What does that do?

hyrisecockpit/run_backend.py
```diff
@@ -0,0 +1,124 @@
+"""Setup script for the starting of the cockpit."""
```
Bouncner: Shebang.

hyrise/Dockerfile
Alexander-Dubrawski: In the beginning, I was using `unzip`. But strangely docker throws an exception when using it in a dockerfile. After some debugging, I still wasn't able to fix this issue so I used p7zip instead and everything was working.

docker-compose.yml
Alexander-Dubrawski: Inside the docker file for the hyrise we are exposing the port 5432 (this port can be used inside the container). In the `compose file` we are publishing the port outside the container. So we need to map the port 5432 to another port on the machine if we want to run more than one hyrise build from this docker file (https://stackoverflow.com/questions/22111060/what-is-the-difference-between-expose-and-publish-in-docker). But I can try to make this process simpler. Maybe it is better to start one hyrise container per default and additional ones manually with `docker run`.

hyrise/Dockerfile
Bouncner: If unzip doesn't work, then we should rather go with `xz` or `bzip2` (I'd guess bzip2 is installed already) to get the compression advantage.
E.g., with `XZ_OPT=-9 tar cJf blub.tar.xz blub` you can use the "extreme" compression of xz. Takes forever to compress, but is small and fast to decompress (ideal for our case).

docker-compose.yml
Bouncner: I missed the part that the Hyrise container uses 5432. So my point doesn't make sense I guess. Please ignore my comment on the ports.

I would vote for a second hyrise instance. AND it would be good to have a clear and well documented way to avoid starting the Hyrise instances and instead pass two external instances to it.

Alexander-Dubrawski: Ok, I will adjust the documentation.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: I'm not sure why I used `ADD`. I changed it to `COPY`. What happens is: 

1. In the first `COPY`, we are copying the files in `hyrisecockpit` inside the docker image. 
2. Inside the workload reader, we need the path to the queries. The path that we are getting is `/usr/local/lib/python3.8/site-packages` instead of `/usr/local/Cockpit`. I`m actually not sure why but I can have a look at it. For the moment we are just copying the workload queries to this location.

Bensk1:
Will look into it as soon as possible.

Bensk1:
Hm, I am feeling a bit bad about removing all that sophisticated Watchtower magic.

hyrise/Dockerfile
```diff
@@ -1,24 +1,48 @@
...
 
-# From now on, we're doing additional stuff that would be done manually otherwise
+# Install packages
 RUN apt-get update
+RUN apt-get -y install wget
+
+# Download binary tables
+RUN mkdir /usr/local/hyrise \
+    && cd /usr/local/hyrise \
+    && wget -O cached_tables.tar.xz ${BINARY_TABLES_URL}
```
Bensk1: What does this currently include? The table's of all benchmarks?

hyrisecockpit/Dockerfile
Bensk1: Maybe add a comment?

Bensk1 (approved):

hyrisecockpit/run_backend.py
```diff
@@ -0,0 +1,124 @@
...
+def create_logger_pipes(component):
+    """Create logger for a component and needed pipes."""
+    logger = logging.getLogger(component)
+    logger.setLevel(logging.INFO)
+    logpipe_stdout = LogPipe(logging.INFO, logger)
+    logpipe_stderr = LogPipe(logging.WARNING, logger)
+    return {"logpipe_stdout": logpipe_stdout, "logpipe_stderr": logpipe_stderr}
+
+
+def main():
```
Bensk1, line 101: Did you use this script in practice yet?

hyrise/Dockerfile
Alexander-Dubrawski: At the moment: 

* tpcc sf 5
* tpcds sf 1
* tpch sf 0.1
* tpch sf 1 

I will try to add job as well.

hyrisecockpit/run_backend.py
Alexander-Dubrawski: Yes. It works pretty well. Moreover, it is a big help in using the cockpit without docker as well.

Bensk1 (approved):

hyrisecockpit/Dockerfile
Alexander-Dubrawski: I added a comment. I think to find a better solution we need to adjust the backend and how we get the current path inside of it.

hyrise/Dockerfile
```diff
@@ -1,24 +1,48 @@
 # https://github.com/hyrise/hyrise/wiki/Docker-Image
 FROM hyrise/opossum-ci:20.04
-LABEL maintainer="upigorsch@me.com"
+
+# Set environment variables
 ENV OPOSSUM_HEADLESS_SETUP=true
+ENV HYRISE_BRANCH bp1920
+ENV THREADS 4
+ENV BINARY_TABLES_URL https://www.dropbox.com/s/yqw3f7ranl79swf/cached_tables.tar.xz?dl=0
```
Bouncner: Isn't `dl=0` the marker that says that the file is not directly provided (no download) but rather an intermediate page is shown? Might work as Dropbox recognizes a non-Browser client, but we should still double check.

hyrise/Dockerfile
```diff
@@ -1,24 +1,48 @@
...
 
-# From now on, we're doing additional stuff that would be done manually otherwise
+# Install packages
 RUN apt-get update
+RUN apt-get -y install wget
+
+# Download binary tables
+RUN mkdir /usr/local/hyrise \
+    && cd /usr/local/hyrise \
+    && wget -O cached_tables.tar.xz ${BINARY_TABLES_URL}
```
Bouncner: Are we now always downloading the files? Or just when the container is created from scratch?
Would it make sense to download it to the client and link it via a docker mount/share?

hyrise/Dockerfile
```diff
@@ -1,24 +1,48 @@
...
+
+# uncompress and unzip binary tables
+RUN cd /usr/local/hyrise \
+    && tar -xf cached_tables.tar.xz \
+    && rm cached_tables.tar.xz
+
+# Install hyrise
 RUN cd /usr/local/ \
-    && git clone https://github.com/hyrise/hyrise.git hyrise-git \ 
+    && git clone https://github.com/hyrise/hyrise.git hyrise-git \
```
Bouncner: How can that work without `--recursive`?

hyrisecockpit/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.8.5-slim
+
+WORKDIR /usr/local/Cockpit
+COPY . .  
+RUN pip3.8 install -r requirements.txt \
+    && pip3.8 install .
```
Bouncner: What is this second `pip install` for?

hyrisecockpit/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+# Python will look in the site-packages directory 
+RUN mv ./hyrisecockpit/workload_generator/workloads /usr/local/lib/python3.8/site-packages/hyrisecockpit/workload_generator/workloads
```
Bouncner: That doesn't look clean. It there another way without the fixed paths?

hyrisecockpit/run_backend.py
```diff
@@ -0,0 +1,124 @@
+"""Script to start all back-end components."""
```
Bouncner, line 3: Shebang.

Bouncner: Also, that's not a comment but a string.

Bouncner (commented):
Some remarks.

Is the Docker help file still up to date? How can I test this all (both docker and setup script) before approving?

.env.example
```diff
@@ -36,10 +36,10 @@ WORKLOAD_PUBSUB_PORT="8003"
...
-STORAGE_HOST="influxdb"
+STORAGE_HOST="127.0.0.1"
 STORAGE_PORT="8086"
 STORAGE_USER="root"
 STORAGE_PASSWORD="root"
 
 FLASK_ENV="development"
-FLASK_DEBUG="True"
\ No newline at end of file
+FLASK_DEBUG="False"
```
Bouncner: Please always end with an empty line (https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

Bouncner (commented):
One small thing.

Alexander-Dubrawski:
> Some remarks.
> 
> Is the Docker help file still up to date? How can I test this all (both docker and setup script) before approving?

How to use the docker setup is described in https://github.com/hyrise/Cockpit/wiki/Docker-Setup. You just need to check out this branch.   
At the moment you could test the script separately by setting up the backend like described in the readme (https://github.com/hyrise/Cockpit). You can then just simply run:

1. `pipenv shell`
2. `cockpit`

hyrise/Dockerfile
Alexander-Dubrawski: We are downloading the files while we are building the image. If we start a container we are using the images and not downloading the files. They are already inside the image.

hyrise/Dockerfile
Alexander-Dubrawski: I changed it to the direct download link.

hyrise/Dockerfile
Alexander-Dubrawski: I'm not sure it just worked. But to be 100% sure I added the `--recursive` argument.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: I tried to figure out why we need the second `pip install` but so far I wasn't able to find a good explanation (I think it has something to do with the setup.py, which takes part in the distribution (builds the hyrisecockpit)). What I can say is that for the moment we need it. Otherwise, we can't execute the entry points and we are getting some import errors.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: Sadly I don't have a better idea at the moment.

hyrisecockpit/run_backend.py
Alexander-Dubrawski: > Also, that's not a comment but a string.

What exactly do you mean by that? It's supposed to be a docstring.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: Maybe this explanation helps a little bit https://stackoverflow.com/questions/57147721/what-does-pip-install-actually-do

hyrisecockpit/Dockerfile
Bouncner: A little bit better would be something like `python -m site` to obtain the path (or other solutions listed here: https://stackoverflow.com/questions/122327/how-do-i-find-the-location-of-my-python-site-packages-directory). Also, do we really need to fix it to `python3.8` rather than just using `python3`? We should write code that is forward portable and the next alpine update might screw us, wouldn't it?

hyrisecockpit/Dockerfile
Bouncner: Can you add a short comment about that? "Second pip required to ... blub"

hyrise/Dockerfile
Bouncner: Do we rebuild the image with every Hyrise update?

Alexander-Dubrawski: At the moment we need to do it manually with `docker image build --no-cache --tag hyrise/bp1920:hyrise ./hyrise` (and change the architecture env in the dockerfile to `core2`). We could adjust the dockerfile in the hyrise bp1929 branch so that a new image version is pushed to docker hub if an update occurs.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: Thanks for the suggestion I will have a look into it :). I agree, if `python3` uses the newest python version it would be better to use it.

hyrise/Dockerfile
Bouncner: If we do it manually (which is fine since we are the developers of Hyrise anyways), would the table data be downloaded each and every time?
And why would we chose core2 if we build on our own machine?

Handling this within the Hyrise branch is not something we should pursue.

hyrisecockpit/Dockerfile
Bouncner: My updated and rather naked Ubuntu tell me that it has Python 3.8.2.

We should double check but having a check for the python version being at least X.Y and afterwards only relying on `python3` sounds better to me than pinning it to `python3.8`.

hyrise/Dockerfile
Alexander-Dubrawski: If we build the image the table data is downloaded. If we start a container not. We only need to chose `core2 ` if we want to push the image to docker hub (so it is working on the most machines). At the moment if we stert the setup we pull the image from docker hub (the tables are in this image included) because it is faster than building the image (but you still have the option to build it manually for your architecture). The disadvantage at the moment is that some person needs to build a new hyrise image if it is updated and push it to docker hub.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: I found a solution. We can define a `MANIFEST.in` file which tells the `setup process` the additional files that are not automatically included in a source distribution.

hyrisecockpit/run_backend.py
```diff
@@ -0,0 +1,127 @@
...
+    def close(self):
+        """Close the write end of the pipe."""
+        close(self.fdWrite)
+
+
+def run_components(loggers_pipes):
+    """Start components in subprocesses."""
+    components = {}
+    logging.info("start cockpit-backend")
+    components["cockpit-backend"] = Popen(
```
Bouncner: That's too much duplicated code. Can we not just pass `cockpit-backend` and that sufficient for the logging info and `Popen`?

hyrise/Dockerfile
Bouncner: Can you document the way we currently handle all the tables file somewhere? So what is created on the fly, what is pushed to Docker Hub, what is stored where ...

hyrise/Dockerfile
```diff
@@ -1,24 +1,63 @@
...
-    && git checkout ${BRANCH:-bp1920} \
-    && sed -i "s/-march=native/-march=${ARCHITECTURE:-native}/g" CMakeLists.txt \
+    && git checkout ${HYRISE_BRANCH} \
+    && sed -i "s/-march=native/-march=${ARCHITECTURE}/g" CMakeLists.txt \
     && ./install_dependencies.sh \
     && mkdir /usr/local/hyrise-git/cmake-build-release \
     && cd /usr/local/hyrise-git/cmake-build-release \
     && CXXFLAGS=-fdiagnostics-color=always cmake -DENABLE_NUMA_SUPPORT=Off -DCMAKE_BUILD_TYPE=Release .. \
-    && make hyriseConsole hyriseServer \
+    && make hyriseServer hyriseBenchmarkTPCH hyriseBenchmarkTPCDS hyriseBenchmarkJoinOrder hyriseBenchmarkTPCC hyriseConsole -j ${THREADS} \
```
Bouncner, line 23: We don't need the benchmarks.

Bouncner: Nevermind!

hyrise/Dockerfile
Alexander-Dubrawski: Ok I wrote a little guide https://github.com/hyrise/Cockpit/wiki/Hyrise-Things

Bouncner (approved):

PR #748, opened 2020-08-26
Simplify development setup

Alexander-Dubrawski:
# Simplify development setup

This PR adds a script that setups the development environment. 
Moreover, there is now a Guide about how to use the development environment: https://github.com/hyrise/Cockpit/wiki/Development-Environment-Setup .

## Todo

- [x] write small guide (include installation hyrise, influx?)
- [x] fix found 5 vulnerabilities (4 low, 1 high) run `npm audit fix` to fix them, or `npm audit` for details
- [x] Change Python version to default (3.8)


Bensk1 (approved):

install_dependencies_development.sh
```diff
@@ -0,0 +1,78 @@
...
+        if brew ls --versions $formula > /dev/null; then
+            echo "$formula already installed"
+            continue
+        elif ! brew install $formula; then
+            echo "Error during brew formula $formula installation."
+            exit 1
+        fi
+    done
+
+    if ! pyenv versions | grep '3.8.5'; then
```
Bouncner: Same discussion here as in #747. Do we really need this specific version? Why not the currently default 3.8.2?

If we need .5, then we should require at least this version (--version >= 3.8.5) but from then on only rely on "any" python3 version to be future-safe.

install_dependencies_development.sh
```diff
@@ -0,0 +1,78 @@
...
+        echo "Install Python 3.8.5 for pyenv."
+        pyenv install 3.8.5
+    fi
+
+    pyenv local 3.8.5
+    python -m pip install pipenv
+    exec "$SHELL"
```
Bouncner: If that is executed by the script, the readme is confusing.

install_dependencies_development.sh
```diff
@@ -0,0 +1,78 @@
...
+    pipenv sync --dev
+    pipenv run pre-commit install
+
+    wget -qO- https://repos.influxdata.com/influxdb.key | sudo apt-key add -
+    source /etc/lsb-release
+    echo "deb https://repos.influxdata.com/${DISTRIB_ID,,} ${DISTRIB_CODENAME} stable" | sudo tee /etc/apt/sources.list.d/influxdb.list
+    sudo apt-get update && sudo apt-get install -y influxdb
+    sudo service influxdb start
+
+    curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
```
Bouncner: Same here again, why 14.x? Do we expect v15 to break something?

install_dependencies_development.sh
Alexander-Dubrawski: I agree the default version is better. I will change it in the backend. That should make things easier.

install_dependencies_development.sh
Alexander-Dubrawski: My mistake. I forgot to remove it from the script.

install_dependencies_development.sh
Alexander-Dubrawski: In the official installation instruction (https://github.com/nodesource/distributions/blob/master/README.md#debinstall), there are just versions listed up to 14.x.

Alexander-Dubrawski: But I can change it to LTS.

Bouncner: LTS would be 12x. I don't know if that backwards step would work. But we could use `Node.js Current (v14.x):` which will always pull the latest version.

I don't know what makes most sense, to be honest. I don't like working with outdated versions, but a change to a new version might crash for whatever reason.

For what do we currently use node after all?

Alexander-Dubrawski: We need node for the frontend.

install_dependencies_development.sh
Bouncner: Ping if done, please.

Alexander-Dubrawski: I changed the python version to 3.8. As a result, we don't need to use `pyenv` anymore which reduces the complexity as well 🎉 .

install_dependencies_development.sh
```diff
@@ -0,0 +1,50 @@
...
+            python3.8-dev \
+            python3.8-distutils \
+            python3-pip \
+            libpq-dev \
+            wget \
+            curl \
+            git 
+
+    python3.8 -m pip install --upgrade pip
+    python3.8 -m pip install pipenv
```
Bouncner: didn't you say that pipenv is no longer required?

Bouncner: Can't we pull that `pipenv` stuff out of the if branches? Looks pretty similar in both cases.

Bouncner (approved):
Small things. Looks good to me.

install_dependencies_development.sh
Alexander-Dubrawski: `pyenv` is no longer required 🙃 .

Alexander-Dubrawski: good point.

PR #750, opened 2020-08-28
Simplify native setup

Alexander-Dubrawski:
This PR simplifies the process of setting up the cockpit in a native environment. For that, it adds an `install_dependencies.sh` script. In comparison to #748 the aim of this PR is to set up the cockpit in a production environment (so no developer environment). It tries to make the process of installing, deleting, and starting the cockpit simpler. You can find a guide on how to install the native setup under: https://github.com/hyrise/Cockpit/wiki/Installation-(Native-Setup) .


install_dependencies.sh
```diff
@@ -36,6 +36,7 @@ python3 -m venv venv
 . venv/bin/activate
 
 echo "Install packages"
+pip3 install wheel
```
Bouncner: For what?

install_dependencies.sh
```diff
@@ -48,5 +49,5 @@ sed -i "s:$PWD/hyrisecockpit/frontend:hyrisecockpit/frontend:g" hyrisecockpit/ru
 
 echo "Install Frontend"
 cd hyrisecockpit/frontend
-npm install --no-optional
+npm install --no-optional && npm audit fix
```
Bouncner, line 59: For what do we actually use `npm`?

hyrisecockpit/Dockerfile
```diff
@@ -6,6 +6,6 @@ COPY . .
 RUN pip3 install -r requirements.txt  \
     && pip3 install .
 EXPOSE 8000
-ENTRYPOINT cockpit
+ENTRYPOINT ["cockpit", "--backend"]
```
Bouncner, line 9: What is that?

install_dependencies.sh
```diff
@@ -0,0 +1,53 @@
...
+            echo "Error during brew formula $formula installation."
+            exit 1
+        fi
+    done
+elif [[ "$unamestr" == 'Linux' ]]; then
+    echo "Installing dependencies (this may take a while)..."
+    curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
+    sudo apt-get install -y nodejs
+    sudo apt-get update
+    sudo apt-get install -y python3
```
Bouncner: Since we are discussing it right now: directly install 3.8?

install_dependencies.sh
```diff
@@ -0,0 +1,53 @@
...
+    sudo apt-get install -y python3-venv
+
+    wget -qO- https://repos.influxdata.com/influxdb.key | sudo apt-key add -
+    source /etc/lsb-release
+    echo "deb https://repos.influxdata.com/${DISTRIB_ID,,} ${DISTRIB_CODENAME} stable" | sudo tee /etc/apt/sources.list.d/influxdb.list
+    sudo apt-get update && sudo apt-get install influxdb
+    sudo service influxdb start
+fi
+
+# Adjust path to frontend (the file will be copied to the python module (at the installation))
```
Bouncner: Please extend this comment significantly. I have no idea why we are doing this.

tests/test_run.py
```diff
@@ -0,0 +1,206 @@
+# type: ignore
+# flake8: noqa
+
+"""Tests for the process manager."""
```
Bouncner, line 4: What is this project manager?

install_dependencies.sh
Alexander-Dubrawski: npm is a software registry. So it manages the javascript packages (like vue,...) that we are using in the frontend. It is something like pip in python. For more information check out: https://docs.npmjs.com/about-npm/.

hyrisecockpit/Dockerfile
Alexander-Dubrawski: The entry point is `/bin/sh -c` and the command that will be executed `cockpit --backend`. So every time docker starts a  container from the image it will run the command `cockpit --backend` first and start the process manager (`hyrisecockpit/run.py`) inside of the container. For more information, you can check out: https://stackoverflow.com/questions/21553353/what-is-the-difference-between-cmd-and-entrypoint-in-a-dockerfile .

tests/test_run.py
Alexander-Dubrawski: Do you mean `process manager` 🙃  ? It starts all the cockpit components (manager, generator, backend, frontend) in subprocesses and manages them. It displays their output and can stop them. The process manager is implemented in `hyrisecockpit/run.py`.

install_dependencies.sh
Alexander-Dubrawski: Wheels are a component of the Python ecosystem that helps to make package installs via pip work. They allow for faster installations and more stability in the package distribution process. So we need it otherwise `pip install` will throw an error.

install_dependencies.sh
Alexander-Dubrawski: I extended the comment. Please give me feedback if it is now more understandable.

install_dependencies.sh
Bouncner: I know wheel, but do we create packages on our own?

`pip install` does not require wheel "in the normal case".

Alexander-Dubrawski: `pip install` uses wheel at the installation process. If we don't install wheel we are getting an error message.

Bouncner: Can you list the installation error? I never needed to install wheel before using pip.

~~Could this be another way to fix it? https://stackoverflow.com/a/54833684~~

Bouncner: Ah ... now I see. We use a setup.py to install the Cockpit? Whatever ... in this case, `wheel` makes sense (I guess).

~~But please still have a look at the solution above and check if this is sufficient.~~

Bouncner: Please add a short comment that we use the setup.py for XYZ and thus need wheel.

Bouncner (approved):
Feel free to merge when the comment is added.

PR #751, opened 2020-08-31
Change python version to 3.8

Alexander-Dubrawski:
This PR changes the required python version from 3.8.5 to 3.8.


Alexander-Dubrawski:
The Python version will be already changed in #748.

PR #752, opened 2020-09-01
Add asynchrone job handler

Alexander-Dubrawski:
This PR is the first of four which aims to refactor the handling of Hyrise jobs (it aims to split #664). Hyrise jobs are interacting directly with the Hyrise to get the meta information and managing the Hyrise settings.  We have three kinds of jobs: 

1. jobs that need to be continuously in the background (for example: get the system information)
2. jobs that will be done asynchronously (for example: loading the tables)
3. jobs that are done synchronously (for example: sending a SQL query over the SQL endpoint)

Both continuous and asynchronous jobs are executed via the apscheduler (https://apscheduler.readthedocs.io/en/stable/). Esspacillay with the asynchronous jobs that's not necessary and increases the complexity (we could just use normal python threads). All three kinds of jobs are scattered in the database and background_scheduler component. The goal of the refactoring is now to have a clear separation for the jobs in the three categories (continuous, asynchronous, and synchronous) and define handlers for that. Moreover, we can stop the continuous jobs while loading/deleting tables (hyrise/hyrise#2180). 

This PR implements the handler for the asynchronous jobs. It uses normal python threads (https://docs.python.org/3/library/threading.html#threading.Thread) for that. 

## ToDo: 

- [x] Adjust `execute_queries_parallel.py` to use Threads instead of processe


hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -0,0 +1,87 @@
+"""Asynchronous job handler."""
```
Bouncner, line 1: shebang

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -0,0 +1,87 @@
...
+from hyrisecockpit.database_manager.cursor import ConnectionFactory
+
+from .job.activate_plugin import activate_plugin as activate_plugin_job
+from .job.deactivate_plugin import deactivate_plugin as deactivate_plugin_job
+from .job.delete_tables import delete_tables as delete_tables_job
+from .job.load_tables import load_tables as load_tables_job
+
+
+class AsynchronousJobHandler:
+    """Handles asynchronous jobs."""
```
Bouncner: Great doc. :D

hyrisecockpit/database_manager/asynchronous_job_handler.py
Bouncner: You moved a lot of code from the `background_scheduler.py` to this file, but I don't see any comment that explains why. Also the PR should give a good reason and not just link another PR's discussion.

Why do we handle certain requests differently? Why isn't everything asynchronous? Why, why, why ...

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -0,0 +1,87 @@
...
+                    self._workload_drivers,
+                ),
+            )
+            job_thread.start()
+            return True
+        else:
+            return False
+
+    def activate_plugin(self, plugin: str) -> bool:
+        """Activate plugin."""
```
Bouncner: What are these doc strings for? They don't help anybody. Are they enforced in some way?

hyrisecockpit/database_manager/asynchronous_job_handler.py
Bouncner: What happens if the methods return `False`? How sets the flags back to "no longer blocked"?

tests/database_manager/test_asynchronous_job_handler.py
```diff
@@ -0,0 +1,225 @@
...
+    def test_successfully_start_delete_tables(
+        self, mock_thread: MagicMock, asynchronous_job_handler: AsynchronousJobHandler
+    ) -> None:
+        """Test successfully start loading tables job."""
+        asynchronous_job_handler._database_blocked.value = False
+        asynchronous_job_handler._connection_factory = MagicMock()
+        mock_job_thread = MagicMock()
+        mock_thread.return_value = mock_job_thread
+        fake_workload_type = "tpch"
+        fake_scale_factor = 1.0
```
Bouncner, line 142: Is data actually loaded during the tests?

Alexander-Dubrawski: No. What happening is that we are mocking the Thread library. So we are not actually starting a Thread instead we are using a mock.  So with `@patch("hyrisecockpit.database_manager.asynchronous_job_handler.Thread")` we are replacing the Thread import in  https://github.com/hyrise/Cockpit/blob/264e329457fcf3c510f5cdefb1ab145e0b08c45b/hyrisecockpit/database_manager/asynchronous_job_handler.py#L4 
with a mock. Let's say the mock object has the id 42. We can then access this mock object in our test. The mock object in this case is an argument to our test function with the name `mock_thread `.  When running the test in the following code we do not actually create but call the mock with the id 42:

https://github.com/hyrise/Cockpit/blob/B/Add_asynchrone_job_handler/hyrisecockpit/database_manager/asynchronous_job_handler.py#L52-L61

Indie the test we can then check if the mock was called with the expected arguments.

hyrisecockpit/database_manager/asynchronous_job_handler.py
Alexander-Dubrawski: We decided to add it to every function. We also have a hook that enforces them. The hope was that we could improve the readability and use it as documentation. But I think it didn't work out so well especially if in some functions are so straight forward that they don't really need them. But in some functions, they can be a help.

hyrisecockpit/database_manager/asynchronous_job_handler.py
Alexander-Dubrawski: I tried to improve the readability of the method by improving the docstring. Why do I need to add a shebang? This file is not directly executed.

hyrisecockpit/database_manager/asynchronous_job_handler.py
Alexander-Dubrawski: I tried to improve the docstrings by using google style docstrings (https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings). I would be very happy about your feedback. If it is now an improvement.

hyrisecockpit/database_manager/asynchronous_job_handler.py
Bouncner: Yeah, that shebang comment was stupid. 😁

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -13,21 +13,44 @@
...
         connection_factory: ConnectionFactory,
         workload_drivers: Dict,
     ):
-        """Initialize asynchronous job manager object."""
+        """Initialize asynchronous job handler object.
+
+        Args:
+            database_blocked: Flag stored in a shared memory map. This flag
+                saves if the Hyrise database is blocked or not.
+            connection_factory: A object to create a connection to the Hyrise
```
Bouncner: an object

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -13,21 +13,44 @@
...
         database_blocked: Value,
         connection_factory: ConnectionFactory,
         workload_drivers: Dict,
     ):
-        """Initialize asynchronous job manager object."""
+        """Initialize asynchronous job handler object.
+
+        Args:
+            database_blocked: Flag stored in a shared memory map. This flag
+                saves if the Hyrise database is blocked or not.
```
Bouncner: s/saves/stores

Bouncner: s/database/instance

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -13,21 +13,44 @@
...
+        """
         self._database_blocked: Value = database_blocked
         self._connection_factory: ConnectionFactory = connection_factory
         self._workload_drivers: Dict = workload_drivers
 
     def load_tables(self, workload_type: str, scalefactor: float) -> bool:
-        """Load tables."""
+        """Start load tabled job.
+
+        This function will check if the database is blocked. If not it will
```
Bouncner: I think a comma is required when a sentence starts with "if". :)

Bouncner (approved):
Mini typos. The code looks good, but I am not too familiar with the whole backend. So I am not really the one to judge. I am trusting you here, that everything is well tested both from an automated test site as well as tested with a complete run through of the cockpit.

PR #749, opened 2020-08-27
Update readme

Alexander-Dubrawski:
This PR updates the readme.

## ToDo 

- [x] translate https://github.com/hyrise/Cockpit/wiki/Nutzeranleitung

PR #753, opened 2020-09-01
Update CODEOWNERS

Alexander-Dubrawski:
Since we are now just three people left 💔  I thought it would be best to adjust the code owners. The reason is that the code owners are getting notification if something changes in their owned code. So we shouldn't spam them with notifications 😢 .


Bouncner:
I did not even know that this `CODEOWNERS` file has an actual function. Wow.

I am fine with the change, but we should at least list everybody of the project on the top of the main readme (I'd use the GitHub names).

Alexander-Dubrawski:
> I did not even know that this `CODEOWNERS` file has an actual function. Wow.
> 
> I am fine with the change, but we should at least list everybody of the project on the top of the main readme (I'd use the GitHub names).

I added the Contributors to the readme in #749 .

Bouncner (approved):

PR #756, opened 2020-09-07
B/add synchronous job handler

Alexander-Dubrawski:
This PR is the third of four which aims to refactor the handling of Hyrise jobs (it aims to split #664). Hyrise jobs are interacting directly with the Hyrise to get the meta information and managing the Hyrise settings.  We have three kinds of jobs: 

1. jobs that need to be continuously in the background (for example: get the system information)
2. jobs that will be done asynchronously (for example: loading the tables)
3. jobs that are done synchronously (for example: sending a SQL query over the SQL endpoint)

Both continuous and asynchronous jobs are executed via the apscheduler (https://apscheduler.readthedocs.io/en/stable/). Esspacillay with the asynchronous jobs that's not necessary and increases the complexity (we could just use normal python threads). All three kinds of jobs are scattered in the database and background_scheduler component. The goal of the refactoring is now to have a clear separation for the jobs in the three categories (continuous, asynchronous, and synchronous) and define handlers for that. Moreover, we can stop the continuous jobs while loading/deleting tables (hyrise/hyrise#2180). 

This PR implements the handler for the synchronous jobs.


hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -105,6 +106,9 @@ def activate_plugin(self, plugin: str) -> bool:
             bool: True if job was successful started, False if database was
                 blocked and job couldn't be started.
         """
+        active_plugins = get_plugins_job(self._connection_factory)
```
Bensk1: `get_plugins_job` sounds like a function that returns a job. `active_plugins` sounds like a variable that holds plugin objects.

hyrisecockpit/database_manager/asynchronous_job_handler.py
```diff
@@ -105,6 +106,9 @@ def activate_plugin(self, plugin: str) -> bool:
...
+        if active_plugins is None or plugin in active_plugins:
```
Bensk1: Does the second part match the documentation above? `False if database was blocked and job couldn't be started`.

hyrisecockpit/database_manager/database.py
```diff
@@ -266,16 +280,11 @@ def delete_data(self, workload: Dict) -> bool:
 
     def activate_plugin(self, plugin: str) -> bool:
         """Activate plugin."""
-        active_plugins = self._get_plugins()
-        if active_plugins is None or plugin in active_plugins:
-            return False
+        # TODO Move in background refactoring to scheduler
```
Bensk1: Is this todo intended?

hyrisecockpit/database_manager/job/execute_sql_query.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class SqlResultInterface(TypedDict):
+    """Interface of a sql result."""
+
+    id: str
+    successful: bool
+    results: List[List[str]]
+    col_names: List[str]
+    error_message: str
```
Bensk1: How's this different from database.py LL 21-30?

hyrisecockpit/database_manager/job/execute_sql_query.py
```diff
@@ -0,0 +1,41 @@
...
+    """Execute sql query on database."""
+    if not database_blocked.value:
+        try:
+            with connection_factory.create_cursor() as cur:
+                cur.execute(query, None)
+                col_names = cur.fetch_column_names()
+                return SqlResultInterface(
+                    id=database_id,
+                    successful=True,
+                    results=[[str(col) for col in row] for row in cur.fetchall()],
```
Bensk1: What is the original datatype of `col` before the cast?

hyrisecockpit/database_manager/job/execute_sql_query.py
```diff
@@ -0,0 +1,41 @@
...
+                    col_names=col_names,
+                    error_message="",
+                )
+        except Error as e:
```
Bensk1: Could we be more specific here? Also, is this error handled anywhere? There should be either proper handling or a crash.

hyrisecockpit/database_manager/job/get_detailed_plugins.py
```diff
@@ -0,0 +1,71 @@
...
+def get_plugins(connection_factory) -> Optional[List[str]]:
+    """Return all currently activated plugins."""
+    try:
+        with connection_factory.create_cursor() as cur:
+            cur.execute(("SELECT name FROM meta_plugins;"), None)
+            rows = cur.fetchall()
+    except (DatabaseError, InterfaceError):
+        return None
+    else:
+        return [row[0].split("Plugin")[0] for row in rows]
```
Bensk1, line 39: A comment explaining the format of the string with an example could increase the readability here.

hyrisecockpit/database_manager/job/get_detailed_plugins.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _get_plugin_setting(connection_factory) -> Plugins:
+    """Return currently set plug-in settings.
```
Bensk1: Method name is singular, comment is plural.

hyrisecockpit/database_manager/job/get_detailed_plugins.py
```diff
@@ -0,0 +1,71 @@
...
+
+    The plug-in settings are obtained from the meta_settings table in the hyrise instance.
+    Inside the hyrise instance not just plug-ins have settings. As a result we need to differentiate
+    the plug-in settings from the other settings. That's why we use the SQL Statement WHERE name LIKE 'Plugin::%';.
+    To get the plug-in name with which we communicate in the cockpit we need to extract it from the name column with
+    row[0].split("::")[1]. So for example from Plugin::Compression::MemorySetting we get Compression.
```
Bensk1, line 49: Good!

hyrisecockpit/database_manager/job/get_detailed_plugins.py
```diff
@@ -0,0 +1,71 @@
...
+def get_detailed_plugins(connection_factory) -> Plugins:
+    """Get all activated plugins with their settings."""
+    if (plugins := get_plugins(connection_factory)) is None:
+        return None
+    if (settings := _get_plugin_setting(connection_factory)) is None:
+        return None
+    return {
+        plugin_name: (settings[plugin_name] if plugin_name in settings.keys() else [])
+        for plugin_name in plugins
+    }
```
Bensk1: I find this hard to read compared to a simple loop. Maybe you get a second opinion from @Bouncner. Is it correct to return an empty list on `else`?

hyrisecockpit/database_manager/job/get_loaded_tables_in_database.py
```diff
@@ -0,0 +1,17 @@
+"""This job return all loaded tables in the database."""
```
Bensk1: returns

hyrisecockpit/database_manager/job/get_loaded_tables_in_database.py
```diff
@@ -0,0 +1,17 @@
...
+
+from typing import Dict, List
+
+from psycopg2 import DatabaseError, InterfaceError
+
+
+def get_loaded_tables_in_database(connection_factory) -> List[Dict[str, str]]:
+    """Return already loaded tables."""
```
Bensk1: Maybe remove the `already`? "Return the database's current tables"?

hyrisecockpit/database_manager/job/get_loaded_tables_in_database.py
```diff
@@ -0,0 +1,17 @@
...
+    try:
+        with connection_factory.create_cursor() as cur:
+            cur.execute("select * from meta_tables;", None)
+            rows = cur.fetchall()
+    except (DatabaseError, InterfaceError):
+        return []
+    else:
+        return [row[0] for row in rows] if rows else []
```
Bensk1, line 33: Are we only extracting the names here?

Bensk1 (changes requested):
First round of comments, more tomorrow.

hyrisecockpit/database_manager/job/get_workload_tables_status.py
```diff
@@ -0,0 +1,39 @@
...
+
+def workload_tables_status(tables_in_database, workload_drivers) -> List:
+    """Get list of all benchmarks which are completely loaded."""
+    workload_tables_status = []
+    for workload_type, driver in workload_drivers.items():
+        for scale_factor in driver.get_scalefactors():
+            loaded_tables = []
+            missing_tables = []
+            workload_tables = driver.get_table_names(scale_factor)
+            loaded = True
```
Bensk1, line 17: Would it be clearer to have this in the `else` branch?

hyrisecockpit/database_manager/job/get_workload_tables_status.py
```diff
@@ -0,0 +1,39 @@
...
+            for table_name, database_representation in workload_tables.items():
+                if database_representation not in tables_in_database:
+                    loaded = False
+                    missing_tables.append(table_name)
+                else:
+                    loaded_tables.append(table_name)
+            table = {
```
Bensk1: naming

hyrisecockpit/database_manager/synchronous_job_handler.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+PluginSetting = TypedDict(
+    "PluginSetting",
+    {"name": str, "value": str, "description": str, "display_name": str},
+)
+Plugins = Optional[Dict[str, List[PluginSetting]]]
+
+
+class SqlResultInterface(TypedDict):
```
Bensk1: This seems familiar

Bensk1 (changes requested):

hyrisecockpit/database_manager/asynchronous_job_handler.py
Alexander-Dubrawski: How about to change `active_plugins` to `active_plugin_names` and `get_plugins_job ` to `get_active_plugin_names` ?

hyrisecockpit/database_manager/asynchronous_job_handler.py
Alexander-Dubrawski: The documentation is missing that the job returns false if the plugin is already active. I will adjust the documentation.

hyrisecockpit/database_manager/job/execute_sql_query.py
Alexander-Dubrawski: It's a `tuple`. I will add the type.

hyrisecockpit/database_manager/job/execute_sql_query.py
Alexander-Dubrawski: I will specify it to handle only database errors.

Alexander-Dubrawski: We are already catching all `psycopg2` errors, so all database errors. Is that solution acceptable? We are catching all the following errors: https://www.psycopg.org/docs/errors.html

hyrisecockpit/database_manager/job/get_detailed_plugins.py
Alexander-Dubrawski: I will refactor this part. It is correct to return an empty list on `else`.

hyrisecockpit/database_manager/job/get_loaded_tables_in_database.py
Alexander-Dubrawski: Yes, we are extracting the table names. I will add a comment and adjust the SQL query to clarify it.

hyrisecockpit/database_manager/job/get_workload_tables_status.py
Alexander-Dubrawski: I'm not sure what exactly you are meaning. We need these variables in both cases `if` and `else` when we are returning `table`.

hyrisecockpit/database_manager/asynchronous_job_handler.py
Bensk1: Better!

hyrisecockpit/database_manager/job/execute_sql_query.py
```diff
@@ -1,36 +1,41 @@
...
             with connection_factory.create_cursor() as cur:
                 cur.execute(query, None)
-                col_names = cur.fetch_column_names()
+                col_names: List[str] = cur.fetch_column_names()
+                results: List[Tuple[Any, ...]] = cur.fetchall()
                 return SqlResultInterface(
                     id=database_id,
                     successful=True,
-                    results=[[str(col) for col in row] for row in cur.fetchall()],
+                    results=[[str(col) for col in row] for row in results],
```
Bensk1, line 34: Still, why do we need the cast to `str` here?

hyrisecockpit/database_manager/job/get_detailed_plugins.py
```diff
@@ -12,25 +12,38 @@
...
     try:
         with connection_factory.create_cursor() as cur:
             cur.execute(("SELECT name FROM meta_plugins;"), None)
-            rows = cur.fetchall()
+            rows: List[Tuple[Any, ...]] = cur.fetchall()
     except (DatabaseError, InterfaceError):
         return None
     else:
+        # >>> print([row[0].split("Plugin")[0] for row in [('ClusteringPlugin',)]])
+        # ['Clustering']
```
Bensk1: Remove before merge

hyrisecockpit/database_manager/job/get_loaded_tables_in_database.py
```diff
@@ -1,17 +1,25 @@
...
         with connection_factory.create_cursor() as cur:
-            cur.execute("select * from meta_tables;", None)
-            rows = cur.fetchall()
+            cur.execute("SELECT table_name from meta_tables;", None)
+            rows: List[Tuple[Any, ...]] = cur.fetchall()
     except (DatabaseError, InterfaceError):
         return []
     else:
+        # >>> print([row[0] for row in [('region_tpch_0_1',), ('customer_tpch_0_1',),]] if rows else [])
+        # ['region_tpch_0_1', 'customer_tpch_0_1']
```
Bensk1: see above

Bensk1 (approved):

hyrisecockpit/database_manager/job/execute_sql_query.py
Bensk1: Fine for me

hyrisecockpit/database_manager/job/get_workload_tables_status.py
Bensk1: I was just referring to `loaded` but never mind, it's fine like that.

hyrisecockpit/database_manager/job/execute_sql_query.py
Alexander-Dubrawski: Inside the flask app, we are serializing and deserializing entities. We are doing that to make the code more robust. For example, if something changes in the response from the manager (database_manager/manager.py) we get an error.  Inside the IPC communication, we are using the JSON form for the messages. So that's why we are serializing and deserializing entities. In the case of a SQL result, we expect it to be a string (because the frontend expects the same). That's why we cast it. The python entities are defined by a model in this case it is defined in: 

https://github.com/hyrise/Cockpit/blob/dev/hyrisecockpit/api/app/sql/model.py

Inside the PR description of #775 I tried to describe how the serializing and deserializing is handled in our flask app.

PR #758, opened 2020-09-09
B/improve backend system tests

Alexander-Dubrawski:
This PR improves system tests. It is now possible to run the system test on the backend running in docker as well if it is running natively. Moreover, the tests are simplified. It is not necessary anymore to set up a second .env. All the environment variables are no defined in `docker.ini` and `native.ini`. 
The output of the containers/ components is logged in files, so it should help for debugging. 
I also wrote a guide on how to use the system tests (https://github.com/hyrise/Cockpit/wiki/System-Tests).


Bensk1:
I just read the guide in the wiki. First, good job, it seems understandable. Second, the introduction of the guide implies that there are two different system tests. Actually, it is the same system test, but the docker version is also starting a Hyrise and an InfluxDB instance, right? If I am correct this should be clarified as two different cases of the same system test.

system_tests/backend_handler.py
```diff
@@ -1,75 +1,55 @@
-"""Wrapper for backend."""
-
-from os import remove
-from signal import SIGINT
-from subprocess import Popen  # nosec
+"""Handler to interact with back-end API."""
```
Bensk1: I don't know if my review comment would cause a lot of further changes but documentation/comments like `"""Handler to interact with back-end API."""` in a file called `backend_handler.py` do not really add value.

system_tests/settings.py
```diff
@@ -1,31 +1,17 @@
...
 STORAGE_USER: str = getenv("STORAGE_USER", "root")
 STORAGE_PASSWORD: str = getenv("STORAGE_PASSWORD", "root")
 
-DATABASE_HOST: str = getenv("DATABASE_HOST", "hyrise")
+DATABASE_HOST: str = getenv("DATABASE_HOST", "localhost")
 DATABASE_PORT: str = getenv("DATABASE_PORT", "5432")
+
+SETUP_TIMEOUT: int = getenv("SETUP_TIMEOUT", 10)  # type: ignore
+DEFAULT_TIME_OUT: int = getenv("DEFAULT_TIME_OUT", 4)  # type: ignore
+REQUEST_TIMEOUT: int = getenv("REQUEST_TIMEOUT", 5)  # type: ignore
```
Bensk1: What is the semantic of this? `getenv()` does return a string, correct? These constants should hold integers but they don't. Also, in the files that include settings.py we are calling, e.g., `int(REQUEST_TIMEOUT)`. Shouldn't we just use `int(getenv(...` to have correct type semantics, avoid the `type: ignore`` and avoid the `int(.` in the includers?

system_tests/test_backend.py
```diff
@@ -1,60 +1,28 @@
-"""Module for system tests."""
+"""Module for backend system tests.
+
+The system test is defined through the following user scenario:
+1. Add a database
+2. Load TPC-H sf 0.1 tables
+3. Start workload generation for TPC-H sf 0.1
```
Bensk1: For consistency `(SF 0.1)`

system_tests/test_backend.py
```diff
@@ -1,60 +1,28 @@
...
-    DATABASE_PORT,
-    STORAGE_HOST,
-    STORAGE_PASSWORD,
-    STORAGE_PORT,
-    STORAGE_USER,
-)
+from system_tests.settings import DATABASE_HOST, DATABASE_PORT, DEFAULT_TIME_OUT
 
-SETUP_TIMEOUT = 3.0
+DEFAULT_TIME_OUT = int(DEFAULT_TIME_OUT)
```
Bensk1: See comment in settings.py

system_tests/test_backend.py
```diff
@@ -145,19 +109,27 @@ def test_returns_available_datasets(self):
...
 
-    def test_initialization_calls(self):
-        """Test initialization calls had no errors."""
-        self.check_stderr()
-
     def test_adds_database(self):
-        """Test adding a database."""
+        """Test adding a database.
+
+        This test is adding a database with the id test_database1.
```
Bensk1, line 113: See my comment regarding comments of self-descriptive code above.

system_tests/test_backend.py
```diff
@@ -145,19 +109,27 @@ def test_returns_available_datasets(self):
...
+        """
         response = self.backend.add_database(
             "test_database1", DATABASE_HOST, DATABASE_PORT
         )
         assert response.status_code == 200  # nosec
 
+    def test_creates_influx_database(self):
+        """Test the corresponding influx database is created."""
+        influx_databases = self.influx_client.get_list_database()
```
Bensk1, line 127: Since influx is a database system itself that could be confusing. Is this a list of Influx database systems or hyrise database systems stored inside influx?

system_tests/test_backend.py
```diff
@@ -166,44 +138,51 @@ def test_added_database_is_in_available_databases(self):
...
-                "number_workers": 8,
+                "number_workers": 2,
                 "dbname": "postgres",
             }
         ]
 
-    def test_loads_default_tables(self):
-        """Test loading default tables."""
+    def test_loads_tpch_sf_0_1_tables(self):
+        """Test loading TPC-H tables with scale factor 0_1."""
```
Bensk1: 0.1

system_tests/test_backend.py
```diff
@@ -166,44 +138,51 @@ def test_added_database_is_in_available_databases(self):
...
-        influx_databases = self.influx_client.get_list_database()
-        assert {"name": "test_database1"} in influx_databases  # nosec
-
-    def test_handles_database_without_errors(self):
-        """Test database handling had no errors."""
-        self.check_stderr()
+        After a wait we check if the tables are completely loaded in the back-end.
+        """
+        sleep(DEFAULT_TIME_OUT)
+        self.check_tables_loading("test_database1", "tpch", 0.1)
```
Bensk1: I would think that `check_tables_loading` means that the tables are being loaded at the moment. The description in L154 sounds to me more like checking if the tables were loaded in the past. 

As a result, I would either change the method name (to `check_tables_loaded`) or the comment above.

system_tests/test_backend.py
```diff
@@ -166,44 +138,51 @@ def test_added_database_is_in_available_databases(self):
...
 
     def test_starts_workload_generator(self):
-        """Test starting of the workload generator."""
-        response = self.backend.start_workload("tpch", 0.1, 300)
+        """Test starting of the workload generator.
+
+        We start the workload generator. It will generate a TPC-H
+        workload with the scale factor 0_1. It will publish 10 tasks
```
Bensk1: 0.1

system_tests/test_backend.py
```diff
@@ -166,44 +138,51 @@ def test_added_database_is_in_available_databases(self):
...
+        """
+        response = self.backend.start_workload("tpch", 0.1, 10)
         assert response.status_code == 200  # nosec
 
     def test_starts_worker_pool(self):
-        """Test starting of the worker pool."""
+        """Test starting of the worker pool.
+
+        This tests starts the worker. As a result the Hyrise instance is
+        now under pressure.
```
Bensk1: Under `load`. A Hyrise instance is never under pressure :)

system_tests/test_backend.py
```diff
@@ -166,44 +138,51 @@ def test_added_database_is_in_available_databases(self):
...
+        """
         response = self.backend.start_workers()
         assert response.status_code == 200  # nosec
 
     def test_returns_historical_metric_values_during_workload_execution(self):
-        """Test responses of the historical metrics."""
-        sleep(4.0)  # wait for query executions
+        """Test responses of the historical metrics.
 
+        After a wait we check the metric data for throughput, latency
```
Bensk1: Minor: After a while

system_tests/test_backend.py
```diff
@@ -216,23 +195,36 @@ def test_returns_historical_metric_values_during_workload_execution(self):
...
+        sleep(DEFAULT_TIME_OUT)
         response = self.backend.activate_plugin("test_database1", "Compression")
 
         assert response.status_code == 423  # nosec
 
     def test_returns_activated_plugins(self):
-        """Test activation of the plugin."""
+        """Test activation of the plug-in.
+
+        THis tests checks if the Compression plug-in is returned correctly
```
Bensk1, line 231: `TH`

Bensk1: Also, the sentence is hard to understand. I think I know what you mean but please rephrase it.

system_tests/test_backend.py
```diff
@@ -254,15 +246,22 @@ def test_returns_activated_plugins(self):
...
         )
         assert response.status_code == 200  # nosec
 
     def test_returns_new_plugin_settings(self):
-        """Test new plugin settings."""
-        sleep(1.0)
+        """Test new plugin settings.
+
+        This test checks if the settings from the Compression plug-in
+        was set correctly.
```
Bensk1: were

system_tests/backend_handler.py
```diff
@@ -1,75 +1,55 @@
...
-        """Initialize backend wrapper."""
-        self.error_file = open("backend_stderr.txt", "w")
-        self.output_file = open("backend_stdout.txt", "w")
+class BackendHandler:
+    """Interacts directly with back-end API.
 
-        self._backend_host = BACKEND_HOST
-        self._backend_port = BACKEND_PORT
+    All requests to the back_end have a time out set. If it
+    fails the test that called the method with the request fails.
```
Bensk1: If the timeout is exceeded, the test...

system_tests/backend_handler.py
```diff
@@ -1,75 +1,55 @@
...
         return get(url, timeout=REQUEST_TIMEOUT)
 
     def get_historical_property(
         self, property: str, startts: int, endts: int, precision: int
     ):
-        """Get historical property."""
+        """Get historical property.
+
+        Sends a get request to a endpoint that returns historical data.
+        The URL needs to be adjusted with the start time stamp, end times stamp,
```
Bensk1: end time

system_tests/backend_handler.py
```diff
@@ -1,75 +1,55 @@
...
+        and the precession.
```
Bensk1: precision

system_tests/backend_handler.py
```diff
@@ -111,52 +105,77 @@ def start_workload(
...
         return get(url, timeout=REQUEST_TIMEOUT)
 
     def set_plugin_settings(
         self, database_id: str, plugin_name: str, setting_name: str, value: str
     ):
-        """Set plugin settings."""
+        """Set plug-in settings.
+
+        This method sends a put request to set the settings for a plug-in that
+        is activated on the database.
```
Bensk1: on the database -> for a specific database?

system_tests/test_docker_backend.py
```diff
@@ -0,0 +1,93 @@
...
+from system_tests.test_backend import TestSystem
+
+SETUP_TIMEOUT = int(SETUP_TIMEOUT)
+TestSystem.__test__ = False  # type: ignore
+
+
+def start_container():
+    """Build and start docker container.
+
+    To make sure that the changes in the back-end code are taking into account,
```
Bensk1: To ensure that the changes of the back-end are taken into account, the back-end's docker container is always rebuilt from scratch.

system_tests/test_docker_backend.py
```diff
@@ -0,0 +1,93 @@
...
+    """Stop and shutdown docker container."""
+    run(["docker-compose", "down"])  # nosec
+
+
+class TestDocker(TestSystem):
+    """Back-end system tests for docker setup.
+
+    This class inherits from the TestSystem class. The TestSystem call defines
+    the user scenario. The TestDocker class implements the method setup_class and
+    teardown_class. The setup_class method is called before the System test (user scenario)
```
Bensk1: The inheritance and which methods are implemented can be easily obtained by looking at the source code.

system_tests/test_native_backend.py
```diff
@@ -0,0 +1,87 @@
...
+    subprocess.send_signal(SIGINT)  # nosec
+    subprocess.wait(timeout=TIME_OUT)
+
+
+class TestNative(TestSystem):
+    """Back-end system tests for a back-end running natively.
+
+    This class inherits from the TestSystem class. The TestSystem call defines
+    the user scenario. The TestNative class implements the method setup_class and
+    teardown_class. The setup_class method is called before the System test (user scenario)
```
Bensk1: see above

system_tests/test_native_backend.py
```diff
@@ -0,0 +1,87 @@
...
+    implement these methods to execute the system test with the a natively running back-end.
+    """
+
+    __test__ = True  # Tell pytest to use TestNative as the test starting point.
+
+    @classmethod
+    def setup_class(cls):
+        """Start back-end running natively.
+
+        This method starts back-end. After a wait time it will initialize
```
Bensk1: starts the back-end

Bensk1 (changes requested):

Alexander-Dubrawski:
> I just read the guide in the wiki. First, good job, it seems understandable. Second, the introduction of the guide implies that there are two different system tests. Actually, it is the same system test, but the docker version is also starting a Hyrise and an InfluxDB instance, right? If I am correct this should be clarified as two different cases of the same system test.

Thanks for the feedback. I adjusted the guide.

system_tests/backend_handler.py
Alexander-Dubrawski: I understand. We are using `pydocstyle` as a hook and this hook forces us by default to put everywhere a docstring. In my opinion, that's not really helpful. I will try to change the `pydocstyle` settings.  But I would like to do it in a separate PR if that ok for you?

Bensk1: Sure!

Bensk1 (approved):

PR #759, opened 2020-09-10
Fix hooks

Alexander-Dubrawski:
Somehow the hooks are not happy again. There are quite a few files where I needed to run the hooks on.
I also changed:
* the python version for the hooks to `python3.8` .
* I removed isort because it behaves (sorts differently)  on Mac OSX and Linux. As a result of that is would be impossible to develop in both platforms. 
* I also changed the action command to use https://github.com/pre-commit/action . So we get a much more helpful message is one hook fails. 

## ToDo:

- [x] write a wiki that explains pipenv  
    - [x] What to consider if a package is installed/uninstalled (update requirements with pipenv run pipenv_to_requirements (https://github.com/gsemet/pipenv-to-requirements))
    - [x] Why and when packages are updated (https://pipenv-fork.readthedocs.io/en/latest/basics.html)
    - [x] Explain dangerous behavior: If I install a new package on branch A (it seems that all packages are updated) the `Pipefile.lock`is updated. If I change to branch B my installation of the packages is not consistent with the  `Pipefile.lock` file on branch B. As a resault the github action can fail. 

- [x] Set a fix package versions in `Pipfile`


requirements-dev.txt
```diff
@@ -11,20 +11,59 @@
...
+pbr==5.5.0 ; python_version >= '2.6'
 pep8-naming==0.11.1
 pipenv-to-requirements==0.9.0
+pipenv==2020.8.13 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
+pluggy==0.13.1 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
 pre-commit==2.7.1
+py==1.9.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
+pycodestyle==2.6.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
 pydocstyle==5.1.1
+pyflakes==2.2.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
```
Bouncner, line 50: Is that autogenerated? Why even care about 2.7?

Bensk1:
Lots of formatting changes, right?

Bensk1 (approved):
Looks good to me, did not check all versions though.

.isort.cfg
```diff
@@ -1,7 +0,0 @@
-[settings]
-# psf/black compatible settings
-include_trailing_comma = True
-force_grid_wrap = 0
-line_length = 88
-multi_line_output = 3
-use_parentheses = True
```
Bouncner, line 7: Line end

.pre-commit-config.yaml
```diff
@@ -29,12 +31,6 @@ repos:
       args: [--autofix, --indent=4, --no-sort-keys]
   - repo: local
     hooks:
-    - id: isort
```
Bouncner, line 32: Why remove?

Pipfile
```diff
@@ -5,24 +5,23 @@ verify_ssl = true
...
-flake8-comprehensions = "*"
-sqlparse = "*"
+pre-commit = "==2.7.1"
+black = "==20.8b1"
+flake8 = "==3.8.3"
+pep8-naming = "==0.11.1"
+rope = "==0.17.0"
+bandit = "==1.6.2"
+pydocstyle = "==5.1.1"
+flake8-bugbear = "==20.1.4"
```
Bouncner, line 15: Why all == instead od >=?

requirements-dev.txt
Alexander-Dubrawski: Yes, it is autogenerated. I'm not sure. I will have a look if we can set some options for the autogeneration.

.pre-commit-config.yaml
Alexander-Dubrawski: I removed `isort` because it behaves (sorts differently) on Mac OSX and Linux. As a result of that is would be impossible to develop in both platforms.

Pipfile
Alexander-Dubrawski: I want to make sure that the hooks are on all machines (included git hub action) the same. Different versions of the hooks behave differently.

PR #764, opened 2020-09-17
Change display of host

Alexander-Dubrawski:
At the comparison, workload, and overview site the hostname was displayed. This was problematic because if the Hyrise instances were running on the same machine no clear visual separation was possible. I changed it so now we see the ID and the host. Please give me feedback if the position of the host is good, or if I should remove it.

Screenshot 2020-09-17 at 18 30 33


Bouncner:
Too many lines in the header.

Looks like this header is also shown on the overview page. Is that right?

Alexander-Dubrawski:
> Too many lines in the header.
> 
> Looks like this header is also shown on the overview page. Is that right?

Yes and the comparison page. We could use the following alternatives: 

Screenshot 2020-09-17 at 20 08 33

The header on the workload page looks at the moment as follows: 

Screenshot 2020-09-17 at 20 01 55

An alternative could be: 

Screenshot 2020-09-17 at 20 09 27

What do you think? In my opinion, it could still be helpful to see the host if the Hyrise instances are running on different machines. For example, if two Hyrise instances are running on the same machine and another one on a different one.

Bouncner:
I am fine with seeing the host name as a hover to ensure that we don't have these additional lines in the comparison view.
So I guess I like your last post + having the hover.

Everything else looks good to me.

Alexander-Dubrawski:
> I am fine with seeing the host name as a hover to ensure that we don't have these additional lines in the comparison view.
> So I guess I like your last post + having the hover.
> 
> Everything else looks good to me.

So with *hover* you mean when I hover with my mouse over the database id that I get to see the host? Something like in this example: https://vuejs.org/v2/guide/#Declarative-Rendering

Screen Recording 2020-09-18 at 12 29 05

@Bouncner what do you think about these changes?

Bensk1:
I am not sure, but I think that's what @Bouncner means, yes.

Bouncner:
I don't like the question mark icon. A simple mouse pointer is fine.

But that's exactly what I wanted. I don't know if that's really what we want in the long run. But for now there at least one way to find the host without screwing up the UI.

Alexander-Dubrawski:
> I don't like the question mark icon. A simple mouse pointer is fine.
> 
> But that's exactly what I wanted. I don't know if that's really what we want in the long run. But for now there at least one way to find the host without screwing up the UI.

I changed the cursor: 

Screen Recording 2020-10-05 at 15 44 10

@Bouncner are you happy with this implementation?

Bouncner:
Fine with me.

Bensk1 (approved):

PR #766, opened 2020-09-18
Fix set plugin setting

Alexander-Dubrawski:
This Pr solves the bug that the frontend sends the display name instead of the setting name to update the plugin settings. I adjusted the backend that it returned the `setting name` as well as the `display name`. This Pr also adjusted the frontend to access the `display name` attribute instead of the `setting name` attribute. 

## Todo:

- [x] When approved run `pre-commit run --all-files`


Bensk1:
> [ ] When approved run `pre-commit run --all-files`

Is this standard procedure for PRs now? Are they failing until you run this command?

hyrisecockpit/api/app/plugin/schema.py
```diff
@@ -11,9 +11,7 @@ class PluginSchema(Schema):
     """Schema of a Plugin."""
 
     name = String(
-        description="Identifier of the Plugin.",
-        required=True,
-        default="CompressionPlugin",
+        description="Identifier of the Plugin.", required=True, default="Compression",
```
Bensk1: Although it was already `CompressionPlugin` before your change, such a specific default seems odd.

hyrisecockpit/api/app/plugin/schema.py
```diff
@@ -46,6 +44,11 @@ class PluginSettingSchema(PluginSettingBaseSchema):
         required=True,
         default="This text describes a setting.",
     )
+    display_name = String(
+        description="Name of the setting that shall be displayed.",
+        required=True,
+        default="Memory Budget (MB)",
```
Bensk1, line 52: see above.

Bensk1: Maybe, I don't understand `default` correctly.

Alexander-Dubrawski:
> > [ ] When approved run `pre-commit run --all-files`
> 
> Is this standard procedure for PRs now? Are they failing until you run this command?

No, it's not standard. Something just went wrong in the last merge with the `dev`. The hooks in the dev were updated, but for one hook the update is missing.

hyrisecockpit/api/app/plugin/schema.py
Alexander-Dubrawski: The default is used in `swagger` to easily test the endpoint. 

Screenshot 2020-09-25 at 16 35 28

hyrisecockpit/api/app/plugin/schema.py
Alexander-Dubrawski: The default here will be used in `swagger` as an example.

Screenshot 2020-09-25 at 16 39 10

Bensk1:
Thanks for clearing that up.

hyrisecockpit/api/app/plugin/schema.py
Bensk1: Okay, never mind.

Bensk1 (approved):

PR #772, opened 2020-09-25
Adjust workload names

### LINKED ISSUE #768: Adjust workload configuration view
At the moment the workloads are displayed as *name* *SF* *scale factor*. So for example TPC-H SF 1. We should change it to  *name* (*SF* *scale factor* ) and change JOB to Join order benchmark.

Alexander-Dubrawski:
# Resolves #768

Before this PR the workload names were displayed as *workload* *SF* *scale factor*. Now they are displayed as *workload* (*SF* *scale factor*). Moreover, the `job` workload is displayed as `Join Order Benchmark`.

**Additional context:**  
Sadly for some reason, the component in the frontend that is displaying the workloads communicated a selected workload very strange: 

1. The frontend asked the backend for the available workloads. 
2. It created a string (for example `tpch SF 1`)
3. When the user selected a workload the workload display component sends the string (for example `tpch SF 1`) to the workload component in the frontend. 
4. This component extracted the scale factor and workload name from this string. 
5. The workload component sends the selected workload extracted from the string somewhere in the frontend (Since it is not documented I don't know where)

Since this logic is a big part of the workload selection componet and the frontend is not documented I tried to solve the problem inside the workload component. As a result, we now have some complex regex expressions. Sadly I have no idea how to solve this problem in a clean way without a knowledge of the logic and Idea behind the approach in the frontend.


hyrisecockpit/frontend/src/meta/workloads.ts
```diff
@@ -1,16 +1,43 @@
...
+    "TPC-H": "tpch",
+    "TPC-DS": "tpcds",
+    "TPC-C": "tpcc",
+    "Join Order Benchmark": "job",
+  };
+  const workload: string = workloadName.split(" (")[0];
+  return workload_type_map[workload];
 }
+
 export function getScaleFactor(workloadName: string): string {
```
Bensk1, line 36: Where is this called and where is `scaleFactor` set?

hyrisecockpit/frontend/src/meta/workloads.ts
```diff
@@ -1,16 +1,43 @@
...
+
 export function getWorkloadName(
   workloadType: string,
   scaleFactor: string
 ): string {
-  return `${workloadType.toUpperCase()} SF ${scaleFactor}`;
+  return `${getDisplayWorkloadName(workloadType)} (SF ${scaleFactor})`;
 }
+
+// The following code is pretty ugly. The workload selector component is working somehow
```
Bensk1: `The following code is pretty ugly.` Do you think that this comment is going to help the reader/developer?

Alexander-Dubrawski: No, it's not. I apologize for the choice of words. I'm just not happy with the `regex` at this part of the code, but I don't have a better solution for now.

Bensk1 (approved):

hyrisecockpit/frontend/src/meta/workloads.ts
```diff
@@ -1,16 +1,43 @@
...
+    "TPC-C": "tpcc",
+    "Join Order Benchmark": "job",
+  };
+  const workload: string = workloadName.split(" (")[0];
+  return workload_type_map[workload];
 }
+
 export function getScaleFactor(workloadName: string): string {
-  return workloadName.split(" ")[2];
+  const regExp = /\(([^)]+)\)/;
```
Bouncner: What does `regExp` do? Can we find a better name?

Alexander-Dubrawski: This regex extracts the scale factor: 

```js
> const regExp = /\(([^)]+)\)/;
> const workloadName = "TPC-C (SF 1)"
> const matches = regExp.exec(workloadName)
> matches
[
  '(SF 1)',
  'SF 1',
  index: 6,
  input: 'TPC-C (SF 1)',
  groups: undefined
]
> matches[1].split(" ")[1];
'1'
```

I could rename it to `reg_extract_scale_factor`.

Bouncner: Yes, that would help.

Bouncner: `regExtractScaleFactor` to keep the style...

hyrisecockpit/frontend/src/meta/workloads.ts
Alexander-Dubrawski: It is called in https://github.com/hyrise/Cockpit/blob/95567463aabe1f9c187bb4059f55e4418650808c/hyrisecockpit/frontend/src/services/workloadService.ts . It is not really set but extracted. So `getScaleFactor` is maybe not the best name.

PR #773, opened 2020-10-09
F/fix disconnected plugin messages

### LINKED ISSUE #771: Plugin messages are disconnected from the graph
**Please describe the bug:**  
Sometimes the plugin messages are disconnected from the graph. 

1. Add a workload
2. Start a workload
3. activate a plugin
4. See an error

Alexander-Dubrawski:
# Resolves #771 

Before this Pr the plugin messages where disconnected:
Screenshot 2020-10-09 at 12 51 37

The problem was the following: Since the backend provides the data (throughput, latency, ...) with a time-delay of three seconds (we only get the data older than three seconds ago) and the plugin log in real-time the plugin log is displayed disconnected. I adjusted the range of the x-axis of the throughput graph to display data older than at least four seconds. So the plugin log is showing if it is activated four seconds ago. 

Screenshot 2020-10-09 at 13 00 27


Alexander-Dubrawski:
Reopened in #774

Bouncner:
?

Alexander-Dubrawski:
> ?

I just mixed something with the base branch up. So I decided that the easiest way is to just open a new one in #774 . 🙈

PR #774, opened 2020-10-09
Adjust x axis timestamp range

### LINKED ISSUE #771: Plugin messages are disconnected from the graph

Alexander-Dubrawski:
# Resolves #771 

Before this Pr the plugin messages where disconnected:
Screenshot 2020-10-09 at 12 51 37

The problem was the following: Since the backend provides the data (throughput, latency, ...) with a time-delay of three seconds (we only get the data older than three seconds ago) and the plugin log in real-time the plugin log is displayed disconnected. I adjusted the range of the x-axis of the throughput graph to display data older than at least four seconds. So the plugin log is showing if it is activated four seconds ago. 

Screenshot 2020-10-09 at 13 00 27


Bouncner:
- The PR text says three seconds, the code four. What's correct?
- Do I get that right that we actively decided to only analyse results that are older than 3/4 seconds? These 3/4 is not the interval of the graph updates? If so, why do neglect such a large time frame?
- I would understand the solution so that we d not present the plugin logs for 3 seconds, but it appears you are rather directly manipulating their timestamps. Do I get that right?

Bensk1:
> * The PR text says three seconds, the code four. What's correct?
> * Do I get that right that we actively decided to only analyse results that are older than 3/4 seconds? These 3/4 is not the interval of the graph updates? If so, why do neglect such a large time frame?
> * I would understand the solution so that we d not present the plugin logs for 3 seconds, but it appears you are rather directly manipulating their timestamps. Do I get that right?

Yes, I would like to know that as well.

Alexander-Dubrawski:
> * The PR text says three seconds, the code four. What's correct?

My mistake I wanted to use three seconds in the code as well. But four would work as well.

> * Do I get that right that we actively decided to only analyze results that are older than 3/4 seconds? These 3/4 is not the interval of the graph updates? If so, why do neglect such a large time frame?

There are two factors why we can't get the live data immediately. The reason for that lies in the backend. First of all, we are using continuous queries inside the influx (https://docs.influxdata.com/influxdb/v1.8/query_language/continuous_queries/). 

https://github.com/hyrise/Cockpit/blob/F/Fix_diconnected_pluginlogs/hyrisecockpit/database_manager/database.py#L126-#L213

For example, the `latency_continuous_query` is running every second and calculates the latency for the last five timestamps. This calculation takes some time so we can't get the live data. Moreover, we have the same problem with other metrics.  For example `detailed query information` (https://github.com/hyrise/Cockpit/blob/F/Fix_diconnected_pluginlogs/hyrisecockpit/api/app/metric/service.py#L88-#L121)
If a worker writes the detailed query information in the `successful_queries` table inside the influx it can happen that we try to read it at time *t* but the worker wouldn't have finished writing it at time *t*. That's why in this case we are using an offset from three seconds. This is kind of a magic number and not an ideal solution.  

Inside the frontend, we are sending a request with a time range from 10 seconds in the past (With the exact number of seconds I'm not 100% sure I need to find the correct code in the frontend) to the live current time. The backend then returns a List with entries for all timestamps. Since t doesn't have data for every time entry in the requested time range (often for the last three seconds or so) it will return empty entries. The graph is then plotting the results resulting in a not completely filled graph.

So, all in all, we have the problem that we can't really tell when exactly the data is available (and with the continuous queries it takes longer to get the live data in comparison to the plugin log). 

> * I would understand the solution so that we d not present the plugin logs for 3 seconds, but it appears you are rather directly manipulating their timestamps. Do I get that right?

What I'm doing is to simply adjust the x-axis range. We still get the same data from the backend but the axis only shows the timestamps until three seconds in the past. An alternative solution would be to adjust the time interval from the request itself to the backend. If you like I could have a look at it, but I'm not sure how difficult it would be.

Bensk1:
Thank you for the detailed description. How long do these continuous queries run on average?

Alexander-Dubrawski:
> Thank you for the detailed description. How long do these continuous queries run on average?

I can't really tell. I would need to measure it.

Bensk1:
> I can't really tell. I would need to measure it.

That's probably not important enough right now, but that's something we should investigate at some point.

Bouncner (approved):

Issue #769, opened 2020-09-25
Data size graph begins from the beginning if we reload the page


Alexander-Dubrawski:
**Please describe the bug:**  
The data size graph starts from the beginning if we reload the page.

**Please describe steps to reproduce:**  
1. Load some tables
2. reload the data size page 

**What is the expected behavior? Please describe:**  
The data size graph should show the date before the reload as well. 

**Screenshot(s):**  

![Screenshot 2020-09-25 at 16 15 15](https://user-images.githubusercontent.com/32880898/94277952-8eff9a00-ff4a-11ea-92e3-59383973e069.png)

PR #775, opened 2020-10-14
Refactor monitor namspace

Alexander-Dubrawski:
First of all, I apologize for the size of this PR (In the future I will try to be at least under 600 lines of change). This Pr refactors the monitor namespace. This was the last part needed to be refactored in the backend. The goal is to make the code easier to read and more robust by adding type safety and unit tests. The tests also have the goal to give a better understanding of the code.  Moreover, after this PR I plan to add a new endpoint to this namespace for the memory footprint with a time interval to solve #769. 

This refactoring implements the same logic as in the other namespaces. It implements the following files: 

* **controller.py** : The controller is responsible for coordinating Flask routes, services, and schemas.
On top, the controller is also responsible to provide the documentation for Swagger.
To provide the documentation for swagger the controller uses the flask_accepts library
in connection with schemas (also used to serialize and deserialize objects) from marshmallow.
The controller uses also models (The model is where the entity itself is defined
in a Python representation)for type safety. All the requests are delegated to the service.
* **model.py**: A model is used for type safety and to represent an entity itself as
a Python representation (object). If we for example getting a response
from the database manager or influxdb we convert this response, which
is often a JSON, to a Python object. These objects are defined by
the models. These objects are often composed of other objects.
* **schema.py**: A schema is responsible for the serialization/deserialization of a given entity.
An entity can be a response from the influxdb or the database manager. This response can
then be deserialized (load) into a python entity (model). On the other hand a python entity (model)
can be serialized (dump) to a JSON-encoded string.
For the deserialization, the post_load decorator is used.
For the schemas, we are using the marshmallow library.
The schemas are also used by the controller for documentation.
* **service.py** : The service is responsible for interacting with the entity. This includes
fetching data from the influx or database manager. The data is then if needed
deserialized into a Python entity (model) by using the corresponding schemas.

The main libraries used are: 
* marshmallow (for schemas) (https://marshmallow.readthedocs.io/en/stable/quickstart.html)
* flask restx (https://flask-restx.readthedocs.io/en/latest/)
* flask_accepts (amongst other things for creating documentation from schemas) (https://github.com/apryor6/flask_accepts)

**note**: Since I wasn't able to adjust the frontend with my limited front-end knowledge 😞  I left the old monitor namespace as well, so that the font-end is still supported.


Bouncner:
Close?

Alexander-Dubrawski:
At the moment the refactoring is not needed. But if we add more features to the backend in the future we can reuse the branch.

PR #776, opened 2020-10-20
Add memory footprint endpoint

### LINKED ISSUE #769: Data size graph begins from the beginning if we reload the page

Alexander-Dubrawski:
# Resolves #769

This PR adds a new `memory_footprint` endpoint to the `/metrics` namespace. This endpoint supports time intervals. Before this PR the memory footprint was calculated by the `/monitor/storage` endpoint which didn't support time intervales. That resulted in the *cut off data size graph*. Moreover, I added a new continuous background job `update_memory_footprint` that runs every second in the background.  It writes every second the current memory footprint inside the `memory_footprint` influx table. The memory footprint is calculated with the following SQL query which is sent to the hyrise database: 

```sql
SELECT SUM(estimated_size_in_bytes) FROM meta_segments;
```

I also added a few docstrings to the `metric` namespace and the `update_storage` job.

## ToDo: 

- [x] check the data size unit in the graph. If needed convert to MB (comparison page)
- [x] adjust tests 
- [x] Refactore update_storage_data.py


.pydocstyle.ini
```diff
@@ -0,0 +1,2 @@
+[pydocstyle]
+ignore = D100,D101,D102,D103,D104,D105,D106,D107,D203,D213,D413,D407,D404,D406,D417
```
Bouncner: Empty line at end. *Short* comment about what all these are?

hyrisecockpit/api/app/metric/controller.py
```diff
@@ -1,4 +1,12 @@
-"""Controllers for metrics."""
+"""Controllers for metric namespace.
+
+The controller is responsible for coordinating Flask routes, services, and schemas.
+On top, the controller is also responsible to provide the documentation for Swagger.
+To provide the documentation for swagger the controller uses the flask_accepts library
+in connection with schemas (also used to serialize and deserialize objects) from marshmallow.
+The controller uses also models (The model is where the entity itself is defined
```
Bouncner, line 7: "for swagger", "(the", "representation) for"

To what service are all requests delegated to?

hyrisecockpit/api/app/metric/model.py
```diff
@@ -1,4 +1,11 @@
-"""Model for metric namespace."""
+"""Model for metric namespace.
```
Bouncner, line 1: Can you be more explicit?

"represent an entity itself"? I don't understand.

To a Python object singular.

This all sounds a lot like deserialization. You should be more explicit why you added this additional layer.

hyrisecockpit/api/app/metric/schema.py
```diff
@@ -1,4 +1,13 @@
-"""Schema for metric namespace."""
+"""Schema for metric namespace.
+
+A schema is responsible for the serialization/deserialization of a given entity.
+An entity can be a response from the influxdb or the database manager. This response can
+then be deserialized (load) into a python entity (model). On the other hand a python entity (model)
```
Bouncner: *P*ython.

hyrisecockpit/api/app/metric/schema.py
```diff
@@ -170,6 +181,41 @@ def make_queue_length(self, data, **kwargs):
...
+
+    timestamp = Integer(
+        title="Queue length",
+        description="Timestamp in nanoseconds since epoch",
+        required=True,
+        example=1585762457000000000,
+    )
+    memory_footprint = Float(
+        title="Memory footprint",
+        description="Size of all tables in database",
```
Bouncner: "Allocated size in bytes"? Size could be understand as tuple count. What is the unit? Does not look like bytes but probably should be.

hyrisecockpit/database_manager/job/update_memory_footprint.py
```diff
@@ -0,0 +1,38 @@
...
+def _get_memory_footprint(connection_factory: ConnectionFactory):
+    time_stamp = time_ns()
+    sql = "SELECT SUM(estimated_size_in_bytes) FROM meta_segments;"
+    memory_footprint = 0.0
+    try:
+        with connection_factory.create_cursor() as cur:
+            cur.execute(sql, None)
+            response = cur.fetchone()
+            # If no tables are loaded the response would be (None,)
+            if response[0]:
```
Bouncner: ` is not None` is more explicit.

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -11,6 +9,23 @@
 
 
 def _create_storage_data_dataframe(meta_segments: DataFrame) -> DataFrame:
+    """Manipulate the data frame.
```
Bouncner: Where is this grouping done?

I don't think such examples are bad, but saying that you group by a column and take a sum would be sufficient, I'd say.

Bouncner: Additionally, the example does not really help is there is no grouping of multiple tuples shown.

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -56,7 +71,10 @@ def _create_storage_data_dataframe(meta_segments: DataFrame) -> DataFrame:
...
 
 def _create_storage_data_dictionary(result: DataFrame) -> StorageDataType:
-    """Sort storage data to dictionary."""
+    """Sort storage data-frame to dictionary.
+
+    For a detailed example please refer to the tests for this job.
+    """
     output: StorageDataType = {}
     grouped = result.reset_index().groupby("table_name")
     for column in grouped.groups:
```
Bouncner: The following code is bad practice. Why group and then iterate over the rows?
Please have a look at Pandas' named aggregations and how we could use them.

Maybe you're doing more than I currently realize and that's not trivial to do, but then the example needs to show exactly that.

hyrisecockpit/frontend/src/services/transformationService.ts
```diff
@@ -213,9 +213,9 @@ function getLatencyData(data: any, primaryKey: string = ""): number[] {
   );
 }
 
-/** calculate memory footprint data */
+/** transform and convert memory footprint data from bytes to MB  */
 function getMemoryFootprint(data: any): number[] {
-  return [getDatabaseMemoryFootprint(data)];
+  return data.map((entry: any) => Math.round((entry["memory_footprint"] / 1_000_000) * 1_000) / 1_000);
```
Bouncner: ` * 1_000) / 1_000` ???

Bouncner (commented):
First look.

The schemata and models small like over engineered classes. But I am not familiar with the code. Please be careful with that stuff. We don't want to make it more complex that necessary.

hyrisecockpit/api/app/metric/controller.py
Alexander-Dubrawski: The service is defined in `hyrisecockpit/api/app/metric/service.py` .

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: I understand. I tried to help to understand the code more. But I think a better solution would be to try to simplify the code by itself. I will try to refactor it.

hyrisecockpit/frontend/src/services/transformationService.ts
Alexander-Dubrawski: That's a technique in javascript to cut off the decimal places. What's happening is: 

```javascript
> num = 1.2345334
> Math.round((num) * 1_000)
1235
> Math.round((num) * 1_000) / 1_000
1.235
```

hyrisecockpit/api/app/metric/model.py
Alexander-Dubrawski: I reworked the docstring. Is it now more understandable?

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: I refactored the code. Therefore I removed pandas completely. Please give me feedback if the code is more readable and if I should add some docstrings.

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: I'm not sure why we used pandas. I removed it and simplified (I hope 🙈 ) the code.

hyrisecockpit/frontend/src/services/transformationService.ts
Alexander-Dubrawski: Sorry that *java script* is so special 🤓 . It seems that there is not really a correct way to round. I found this interesting post:

https://stackoverflow.com/questions/11832914/round-to-at-most-2-decimal-places-only-if-necessary

Bouncner: JavaScript sucks.

Any how, please add a comment and I still don't see why we need to do this `* 1000 / 1000`. Can you give me an example? If when want to have three digits rounded, why not `(bytes / 1_000) * 1000`?

hyrisecockpit/api/app/metric/controller.py
```diff
@@ -109,6 +119,25 @@ def get(self) -> List[QueueLength]:
         return MetricService.get_queue_length(time_interval)
 
 
+@api.route("/memory_footprint")
+class MemoryFootprintController(Resource):
+    @accepts(
+        dict(name="startts", type=int),  # noqa
+        dict(name="endts", type=int),  # noqa
+        dict(name="precision", type=int),  # noqa
```
Bensk1, line 127: Is this endpoint convention to not use `_` to separate something like `start_ts`?

hyrisecockpit/database_manager/job/update_memory_footprint.py
```diff
@@ -0,0 +1,38 @@
...
+    memory_footprint = 0.0
+    try:
+        with connection_factory.create_cursor() as cur:
+            cur.execute(sql, None)
+            response = cur.fetchone()
+            # If no tables are loaded the response would be (None,)
+            if response[0] is not None:
+                memory_footprint = float(response[0])
+            else:
+                memory_footprint = 0.0
```
Bensk1: This branch does not change `memory_footprint` and could be removed.

hyrisecockpit/database_manager/job/update_memory_footprint.py
```diff
@@ -0,0 +1,38 @@
...
+    except (DatabaseError, InterfaceError):
+        memory_footprint = 0.0
```
Bensk1, line 22: In this case, I believe that using NaN would be cleaner but I don't know how this would be displayed in the frontend.

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -1,93 +1,93 @@
...
+    vector_compression_type = new_encoding["compression"][0]
+    for encoding in encodings:
+        if encoding["name"] == new_encoding["name"]:
+            encoding["amount"] += 1
+            if vector_compression_type not in encoding["compression"]:
+                encoding["compression"].append(vector_compression_type)
+            encoding_exists = True
+            break
+    if not encoding_exists:
+        encodings.append(new_encoding)
```
Bensk1: What's the meaning of this `if`? It reads like "if this encoding does not exist append it to the list of encodings".

hyrisecockpit/frontend/src/services/transformationService.ts
Alexander-Dubrawski: We need to round the numbers for the graph representation. 

Screenshot 2020-11-10 at 10 27 24

If we are using `(bytes / 1_000) * 1000` we are not cutting off the digits. 

```js
> num = 1.2345334
> (num / 1_000) * 1000
1.2345334
```

Bouncner: Sorry, I meant `Math.Round(num / 1_000) * 1000`.

hyrisecockpit/api/app/metric/controller.py
Alexander-Dubrawski: Yes. To change it would be quite some work because we would need to change it at many other endpoints to keep everything uniformly. I can create an issue for that if you want.

hyrisecockpit/database_manager/job/update_memory_footprint.py
Alexander-Dubrawski: I think (but I'm not 100% sure) the graph in the frontend should just ignore the point with the NaN value. So it wouldn't be completely connected.

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: That's correct. Should I change and use the variable `not_existing_encoding` ?

hyrisecockpit/frontend/src/services/transformationService.ts
Alexander-Dubrawski: That makes sense 🙃 . I will change it.

hyrisecockpit/api/app/metric/model.py
Bouncner: It doesn't really help me. Where can I see such a "model"?

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -11,6 +9,23 @@
...
+        would be transformed to the following data-frame:
+        table_name | column_name | estimated_size_in_bytes | encoding_type                                                                 | vector_compression_type  | column_data_type
+        -----------+-------------+-------------------------+-------------------------------------------------------------------------------+--------------------------+-----------------
+        customer   |   c_custkey |                   90096 | [{'name': 'Dictionary', 'amount': 1, 'compression: ["FixedSize2ByteAligned"]} | ["FixedSize2ByteAligned"]| int
+                   |   c_address |                  975528 | [{'name': 'Dictionary', 'amount': 1, 'compression: ["FixedSize2ByteAligned"]} | ["FixedSize2ByteAligned"]| string
+        lineitem   |  l_orderkey |                  196558 | [{'name': 'Dictionary', 'amount': 1, 'compression: ["FixedSize2ByteAligned"]} | ["FixedSize2ByteAligned"]| int
+        It is important to note that the data-frame is grouped by the table_name column.
+    """
     meta_segments.set_index(
```
Bouncner: Could you please send me an exemplary CSV file of that initial data frame?

hyrisecockpit/api/app/metric/model.py
Alexander-Dubrawski: The models are defined in the file. And they are for example used in the `hyrisecockpit/api/app/metric/schema.py` in the  @post_load methods (https://marshmallow.readthedocs.io/en/stable/quickstart.html#deserializing-objects-loading). The schemas are used in the service for deserializing. So if a response reaches the service from the `database_manager` (running in a separate process) the post_load method is called in the Shema which creates the model.

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: This code is not up to date 🙃 . I removed the dataframes.

hyrisecockpit/api/app/metric/controller.py
Bensk1: No, it's not necessary, thanks.

hyrisecockpit/database_manager/job/update_memory_footprint.py
Bensk1: Semantically, this would be "more" correct. However, DatabaseErrors or InterfaceErrors should not occur too often anyway.

hyrisecockpit/database_manager/job/update_storage_data.py
Bensk1: Now that I see it again, it's fine.

hyrisecockpit/database_manager/job/update_memory_footprint.py
Alexander-Dubrawski: So is it ok if we are leaving it that way? 🙃

Bensk1: Yes

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -1,93 +1,127 @@
...
+    """
+    This method checks if an encoding entry for this encoding type exists.
+    If so it edits the entry by updating the amount and if needed adding
+    a new vector_compression_type to the compression list.
+    """
+    for encoding in encodings:
+        if encoding_type == encoding["name"]:
+            encoding["amount"] += amount
+            # If the encoding exists we need to check if the vector compression type
+            # is in the encoding entry. A encoding entry can have multiple
```
Bouncner: "If the encoding existsCOMMA". 

"*An* encoding entry".

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -1,93 +1,127 @@
...
+            # vector compression types.
+            if vector_compression_type not in encoding["compression"]:
+                encoding["compression"].append(vector_compression_type)
+            return
+    encodings.append(
```
Bouncner, line 27: That's an really aweful style. Not your fault, but sometimes I absolutely don't like this coding style.

Bouncner: Well, rather formatting than coding style.

hyrisecockpit/database_manager/job/update_storage_data.py
```diff
@@ -1,93 +1,127 @@
...
     database_blocked,
     connection_factory,
     storage_connection_factory: StorageConnectionFactory,
 ) -> None:
-    """Get storage data from the database."""
+    """Get the storage information from hyrise.
+
+    This function requests the storage information via SQL.
+    To keep the returned result small we use the group by statement.
+    The amount aggregation returns how often the tuple exists and so
```
Bouncner: Please don't use "amount" for easily countable things. You don't need to fix it in the UI (the JSON uses it as well), but please use "*_count" or occurrences here.

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: Do you have any suggestions for improving this code? Sets are not working well with JSON.

hyrisecockpit/database_manager/job/update_storage_data.py
Alexander-Dubrawski: I can adjust the JSON as well and change `amount` to `occurrences `.

hyrisecockpit/database_manager/job/update_storage_data.py
Bouncner: No. :-/

Bouncner (approved):

PR #777, opened 2020-10-26
Fix segment configuration heatmap

Alexander-Dubrawski:
This PR fixes the problem with the Segment Configuration.

Before this PR  the heatmap looked as followed:

IMAGE 2020-09-18 18_21_24

To solve this Issue I adjusted the `monitor/segment_configuration` endpoint in the backend so that it returns the data in a form that is more suitable to build the heatmap. The *encoding type* and the *order type* mapped to an integer. The type can be found in a `mode_mapping` list inside the response to get the real value. For example 0 maps to Dictionary. This way the frontend can use the mapping List for the heat bar in the heatmap and doesn't need to do some extra calculations. 
As a result, I simplified the corresponding code inside the frontend.

Moreover, we can also use the heatmap to show the order mode of the columns. 

For that simply press this button: 

Screenshot 2020-10-26 at 17 58 19

## ToDo:

- [x] lint all files
- [x] Adjust e2e tests
- [x] Remove Pydocstyle hook (inseperate PR?)
- [x] Rename Y-Axis to `Chunk ID` and remove `No.`
- [x] Try to adjust the CSS properties (to handle the overflow) (have a look at https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Sizing_items_in_CSS)
- [x] Create Issue for integrating *the sorting information into the encoding graph*
- [x] ~~Set flexible trick angle in heatmap~~(moved to issue)
- [x] Refwork docstring in update_segment_configuration.py
- [x] Check y-axis (Why are the columns shown in four steps )


Bouncner:
Can you show the screenshots of the new UI?

Alexander-Dubrawski:
> Can you show the screenshots of the new UI?

Screenshot 2020-10-26 at 18 09 29


I could adjust the x-axis as shown in the following example: 

Screenshot 2020-10-26 at 18 36 42

Bouncner:
Since there is no issue links: was that fix about changing x and y?

Can you name the y-axis "Chunk ID" and skip the "No."?
Is that mock data?

Concerning the axis, we could do something that starts at 30 degrees and goes up to 90 once there are more than 20 columns.

Looks good so far. Do you know how hard it would be to integrate the sorting information into the encoding graph? E.g. an arrow in each cell to denote "sorted"?

.pydocstyle.ini
```diff
@@ -0,0 +1,2 @@
+[pydocstyle]
+ignore = D100,D101,D102,D103,D104,D105,D106,D107,D203,D213
```
Bouncner: empty line to end, what are these?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
+from typing import Dict, List, Tuple
 
 from hyrisecockpit.database_manager.cursor import StorageConnectionFactory
-from hyrisecockpit.database_manager.job.sql_to_data_frame import sql_to_data_frame
+from hyrisecockpit.database_manager.cursor import ConnectionFactory
+
+from psycopg2 import DatabaseError, InterfaceError
+
+
+def _execute_sql(sql: str, connection_factory: ConnectionFactory):
```
Bouncner: This looks like a function that should not be here (as well). I'd guess we have the same functionality already somewhere else?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
-from typing import Dict
-
-from pandas import DataFrame
+from typing import Dict, List, Tuple
 
 from hyrisecockpit.database_manager.cursor import StorageConnectionFactory
-from hyrisecockpit.database_manager.job.sql_to_data_frame import sql_to_data_frame
+from hyrisecockpit.database_manager.cursor import ConnectionFactory
+
+from psycopg2 import DatabaseError, InterfaceError
```
Bouncner: Don't encapsulate the psycopg2 dependency somehow within a single file?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
+            mode_id_mapping.append(mode)
+        formatted_results[table_name][column_name][chunk_id] = mode_id_mapping.index(
+            mode
+        )
 
-    for table_name in grouped_tables.groups:
-        segment_configuration_data[table_name] = {}
-        table = grouped_tables.get_group(table_name)
-        grouped_columns = table.reset_index().groupby("chunk_id")
+    for table_name, column in formatted_results.items():
```
Bouncner: Isn't the first loop `for table_name, column_id ...` and the second `for _, chunk_id ...`?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
+        for column_name, mode in column.items():
+            formatted_mode = [mode[key] for key in sorted(mode)]
```
Bouncner: Please don't sort repeatedly.

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
+            formatted_results[table_name][column_name] = formatted_mode
```
Bouncner: I can't follow. First, you create a dicts of dicts of dicts and now you overwrite the third layer with a list?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,109 @@
...
+                                    encoding_type
+                                    FROM meta_segments;"""
+    sql_segments_order: str = """SELECT
+                                meta_chunk_sort_orders.table_name,
+                                meta_segments.column_name,
+                                meta_chunk_sort_orders.chunk_id,
+                                meta_chunk_sort_orders.order_mode
+                                FROM meta_chunk_sort_orders
+                                JOIN meta_segments
+                                    ON meta_segments.table_name == meta_chunk_sort_orders.table_name
```
Bouncner: That `==` is no valid SQL.

Bouncner (commented):
Stopping for now to get your responses.

Alexander-Dubrawski:
> Since there is no issue links: was that fix about changing x and y?

The fix goes a little bit deeper. In general, the frontend didn't read the data correctly and the backend didn't return the correct data. As a result, the data wasn't plotted correctly. 

> Can you name the y-axis "Chunk ID" and skip the "No."?
> Is that mock data?

Yes, I can adjust the naming. Data is not mocked.

> Concerning the axis, we could do something that starts at 30 degrees and goes up to 90 once there are more than 20 columns.

The Problem at the moment is that we are using DOM Objects that are kind of fixed and I wasn't able to find out how to manipulate their CSS properties. If we use an angle of 90 degrees the content would overflow and be cut off. But I will have a look again where I can find these properties. 

> Looks good so far. Do you know how hard it would be to integrate the sorting information into the encoding graph? E.g. an arrow in each cell to denote "sorted"?

I think in theory that should be possible but sadly I have not the knowledge to implement it. I could create an Issue for that.

.pydocstyle.ini
Alexander-Dubrawski: The Pydocstyle hook can make some problems. If I, for example, ignore one rule then it can happen that as a result, I need to ignore multiple rules. Since I'm not really a fan of this hook I would like to remove it completely.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: Actually, you are right. Normally all the jobs are executing a SQL query that way.  If you want I can put this functionality in a job called `execute_sql.py`.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: I'm not sure what exactly you mean. Can you be more specific 🙃  ? We only import the Error Type and not the hole model.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: We extracting the column name from the table so column id wouldn't be correct. But I agree that chunk_id in the secound loop would be better.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: I want to make sure that the indexes of the list are in the right order. If we would have extraceted for example the following result: 

```python
formatted_results = {
    "lineitem_tpch_0_1": {
        "l_orderkey": {
            "1": 0,
            "0": 1,
            "2": 0,
        },
        "l_partkey": {
            "1": 1,
            "0": 1,
            "2": 1,
        },
    },
}

mode_id_mapping = ["Dictionary", "LZ4"]
```

we would need to sort the keys so that the indexes of the list are in the correct order. We are doing all of this so that the frontend can build the heatmap. The heatmap needs integer values (for the z-axis) and can not work with strings. Moreover, the frontend needs a mapping list to blot the heat bar. So it knows that for example in a list ["Dictionary", "LZ4"] all values with 0 get the color green all with the value 1 yellow and so on. In the heat bar for green would be Dictionary displayed and so on.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: Please refer to the comment above.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: Yes, you are right.

.pydocstyle.ini
Bouncner: What is a hook? What is "this hook"?

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: That shouldn't be a job, that should be a function in a database handler/object/class. We shouldn't define such a basic function again and again.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: Belongs to the discussion above. If we decide to go for another library (might happen, e.g., to get asynchronous queries or whatever), a lot of places need to be touched. This should be encapsulated.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: It's not just the naming. This looks wrong to me.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: What would be the problem is the list is not sorted?

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: The comment above doesn't help me.

Bouncner:
> The fix goes a little bit deeper. In general, the frontend didn't read the data correctly and the backend didn't return the correct data. As a result, the data wasn't plotted correctly

Ok, I see.
 
> > Is that mock data? 
> Yes, I can adjust the naming. Data is not mocked.

Then why is just every forth chunk shown? Or do we show them all but have only every forth shown in the in axis?
I think we definitely need a thin border in the table.

> The Problem at the moment is that we are using DOM Objects that are kind of fixed and I wasn't able to find out how to manipulate their CSS properties. If we use an angle of 90 degrees the content would overflow and be cut off. But I will have a look again where I can find these properties.

That should be straightforward. Nothing important for now, but if you cannot find it easily, please create a ticket.
Can you please check out some TPC-H and DS tables on a laptop screen? If 45 degrees works, I am fine with it for now.

> > Looks good so far. Do you know how hard it would be to integrate the sorting information into the encoding graph? E.g. an arrow in each cell to denote "sorted"?
> 
> I think in theory that should be possible but sadly I have not the knowledge to implement it. I could create an Issue for that.

Yes, please create an issue.

.pydocstyle.ini
Alexander-Dubrawski: We are running these hooks in every commit. For more information please refer to https://github.com/hyrise/Cockpit/wiki/Development-Environment-Setup#Hooks .

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: The functionality is defined in the `connection` that we get from the `connection factory`:

https://github.com/hyrise/Cockpit/blob/dev/hyrisecockpit/database_manager/cursor.py

We still need to use the context manager, so that the cursor is closed properly.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: We handle the errors differently across the backend. We could try to move the Error handling insider the cursor, but I would prefer to do it in a separate PR or create an issue for the moment.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: I will rework it.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: Then in the heat map, we would display the wrong information for the wrong chunk. The index of the list defines the chunk id. Like I said before this endpoint is adjusted to work with the plotting of the heatmap in the frontend. The endpoint by itself and not in connection with the frontend might seem a little bit strange.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: If we can guarantee that the chunk id is ascending in the response of the Hyrise I can build the list directly and we don't need to convert a dict to a list. But the way how it is implemented now we can be sure that the encoding and order is indexed correctly to the chunks.

Bensk1: You can guarantee that by using `ORDER BY`, can't you?

Alexander-Dubrawski: That's a good point. `ORDER BY` should work as well. I will change it.

Alexander-Dubrawski:
> Ok, I see.
> 
> > > Is that mock data?
> > > Yes, I can adjust the naming. Data is not mocked.
> 
> Then why is just every forth chunk shown? Or do we show them all but have only every forth shown in the in axis?
> I think we definitely need a thin border in the table.

that's a good point. I will look into it.

hyrisecockpit/database_manager/job/update_segment_configuration.py
Bouncner: I don't see anything somewhat related to the segment configuration view. So why should we have that method here? If we move it, we could also get rid of the psyco imports.

hyrisecockpit/frontend/src/meta/metrics.ts
```diff
@@ -215,7 +215,7 @@ const metricsChartConfiguration: Record<Metric, ChartConfiguration> = {
   segmentConfiguration: {
     title: "Segment Configurations",
     xaxis: "Segments",
-    yaxis: "",
+    yaxis: "Chunks ID",
```
Bouncner: Chunk ID.

hyrisecockpit/frontend/tests/e2e/specs/metrics/access.spec.ts
```diff
@@ -13,7 +13,7 @@ import {
 
 const metric = "access";
 const request = "chunks";
-const layout = { xtitle: "Columns", ytitle: "Chunks" };
+const layout = { xtitle: "Columns", ytitle: "Chunks ID" };
```
Bouncner: Chunk ID

hyrisecockpit/frontend/src/components/charts/Heatmap.vue
```diff
@@ -99,6 +101,8 @@ function useHeatMapConfiguration(
       },
       yaxis: {
         rangemode: "tozero",
+        tickformat: ",d",
```
Bensk1, line 108: Does that mean "comma separated decimal"? Maybe add a comment then.

hyrisecockpit/frontend/src/components/metrics/SegmentConfiguration.vue
```diff
@@ -201,7 +200,7 @@ export default defineComponent({
       ...selection,
       hoverTemplate: computed(
         () =>
-          `<b>segment: %{text.column}</b> <br>${selection.selectedType.value}: %{text.value} <extra></extra>`
+          `<b>segment: %{x}</b> <br>${selection.selectedType.value}: %{text} <extra></extra>`
```
Bensk1: What is `x`?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
+def _format_results(results: List[Tuple]) -> Dict:
+    """Format psycopg2 cursor results.
 
-def _create_dictionary(
-    segments_configuration: DataFrame, configuration_type: str
-) -> Dict:  # TODO refactoring
-    segment_configuration_data: Dict = {}
-    grouped_tables = segments_configuration.reset_index().groupby("table_name")
+    This function iterates over each row and creates a dictionary where
+    the keys are the tables names. For every table name the value is a
```
Bensk1: `table names` (preferred) or `tables' names`

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
+    dictionary where the keys are the column names. For every column name
+    the value is a list where the indices are the chunk_id and the value is
+    the mode. The mode can be for example "Dictionary",
```
Bensk1: Instead of `mode` something like `...name of the configuration`?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
+    We can build this list with a loop since the chunks are
```
Bensk1: This chunk order is a hard requirement, isn't it? In that case, I would place this comment right at the loop to avoid it getting overseen in the middle of this comment block. Also, should we (debug-)assert that ordering?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
+    in ascending order. The mode is represented by an integer.
+    The integer is the index of the mode in the mode_id_mapping list.
+    We use this list to make it easier in the frontend to build the headmap.
```
Bensk1: The last sentence (`We use this list...`) does not add any value and could be omitted.

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
     database_blocked,
-    connection_factory,
+    connection_factory: ConnectionFactory,
     storage_connection_factory: StorageConnectionFactory,
 ) -> None:
-    """Update segment configuration data for database instance."""
+    """Update segment configuration data for database instance.
+
+    First the needed information is extracted from the Hyrise. After that the raw SQL results
+    are formatted to a dictionary they are written to the influx.
```
Bensk1: `to a dictionary and are written to InfluxDB`?

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,104 @@
...
+    The table meta_chunk_sort_orders only has the columns table_name, chunk_id, column_id and order_mode.
+    To be able to assign the chunks to the column name we need to join the meta_chunk_sort_orders with the
+    meta_segments table.
```
Bensk1: Not sure whether this is necessary or directly visible from the SQL statements.

Alexander-Dubrawski:
@Bensk1 @Bouncner here is a screenshot of the implementation, what do you think? An angle of 45 degrees seems to work for the moment. 

Screenshot 2020-11-10 at 14 17 33

Bouncner:
Angle looks good, but I need to see it with a margin between the cells to make my final verdict.

Alexander-Dubrawski:
> Angle looks good, but I need to see it with a margin between the cells to make my final verdict.

What exactly do you mean by the margin between the calls? do you mean the margin between the xlabels?

Bouncner:
White space between the cells. I cannot find the review comment, but I mentioned it earlier. It shouldn't be a green large rectable, but many green ones. Otherwise nobody understands why the chunk ID show up  like "0, 4, 8, 12".

hyrisecockpit/frontend/src/components/metrics/SegmentConfiguration.vue
Alexander-Dubrawski: It's the x value.

Alexander-Dubrawski:
> White space between the cells. I cannot find the review comment, but I mentioned it earlier. It shouldn't be a green large rectable, but many green ones. Otherwise nobody understands why the chunk ID show up like "0, 4, 8, 12".

Ok, I will look into it.

Alexander-Dubrawski:
> White space between the cells

@Bouncner what do you think about the following heatmap: 

Screenshot 2020-11-10 at 21 06 13

Bensk1:
Fine for me!

hyrisecockpit/frontend/src/components/metrics/SegmentConfiguration.vue
Alexander-Dubrawski: So for example the name of the segment.

Bouncner:
Looks superb. BUT I still would like to see the result for TPC-H SF 1 and table lineitem after the compression plugin was activated. Because that's basically what this whole heatmap is for.

Alexander-Dubrawski:
> Looks superb. BUT I still would like to see the result for TPC-H SF 1 and table lineitem after the compression plugin was activated. Because that's basically what this whole heatmap is for.

That's how it looks on my MacBook (2560 x 1600)

Screenshot 2020-11-11 at 17 29 00

And that's how it looks on my 4k Monitor: 

Screenshot 2020-11-11 at 17 29 40


I don't like that we cut off The encoding at the moment (Diction..).  And which plugin/settings can I use to get some other encoding into the table?

Bouncner:
The compression plugin.

Alexander-Dubrawski:
> The compression plugin.

How do you like it? Beautiful, isn't it? 🙃 

Screenshot 2020-11-14 at 10 17 22

Bouncner:
😍

Bensk1:
That looks great, thanks!

hyrisecockpit/database_manager/job/update_segment_configuration.py
```diff
@@ -1,77 +1,94 @@
...
-    segments_configuration: DataFrame, configuration_type: str
-) -> Dict:  # TODO refactoring
-    segment_configuration_data: Dict = {}
-    grouped_tables = segments_configuration.reset_index().groupby("table_name")
+    This function iterates over each row and creates a dictionary where
+    the keys are the table names. For every table name the value is a
+    dictionary where the keys are the column names. For every column name
+    the value is a list where the indices are the chunk_id and the value is
+    the name of the configuration. The name of the configuration can be for example "Dictionary",
+    The name of the configuration is represented by an integer.
```
Bouncner: Not configuration, “encoding”.
What if Chunk ID 0 does not exist?

Bouncner (changes requested):

hyrisecockpit/database_manager/job/update_segment_configuration.py
Alexander-Dubrawski: We use the same function for getting the order. That's why I used `configuration `. Can it happen that the `meta_segments` tables are missing some Chunk IDs? If some Chunk Ids are missing (and so the chunks itself) nothing would break, but we would assign the chunks the wrong ID.

Bouncner: Yes, chunks might be deleted. Since we are going to have "self-driving" plugins that regularly delete chunks, we should consider fixing this.
@Bensk1 : shall we open an issue? It's important, but the deadline is approaching.

Bouncner: And concerning "configuration". Could we use "property"? We use the term configuration in the index selection, compression selection and elsewhere, where is describes the whole system. So it's confusing to use it here as well.

Alexander-Dubrawski: I can try to fix it.

Alexander-Dubrawski: I wrote a fix for this problem. Which plugin deletes chunks ? I still need to make sure that this fix is working with the frontend as well (In theory it should work).

Bouncner (approved):
Just the missing implementation of the aggregation query. As soon as this is implemented and tested, we can merge.

Alexander-Dubrawski:
> Just the missing implementation of the aggregation query. As soon as this is implemented and tested, we can merge.

Do you mean the query for the #776 PR? 🙃

PR #782, opened 2020-11-09
Remove pydocstyle hook

Alexander-Dubrawski:
This PR removes the PyDocstyle pre-commit hook. The PyDocstyle hook was run every commit. Its main task was to check all docstrings. By default, it required a docstring in every module, class, and method. In my opinion, it does not make sense to force to write docstrings everywhere. When excluding some docstring rules there were strange side effects. Sometimes when excluding one rule I needed to exclude another one because they were depending on each other. All in all, I do not really see the benefit of this hook for our project. This hook would make sense if we plan to create documentation on the base of the docstring (for example with python sphinx). Since we are not doing this I do not really see the benefit. It is more restricting than helpful.


Bouncner:
Hmm, not sure what I think about that.

Bensk1:
In the past, some docstrings clearly only existed because we were forced to have docstrings for everything. Therefore, I think these changes make sense.

Bensk1 (approved):

